PR #2, opened 2026-05-12
Submodules

aaishwarymishra:
This pull request refactors the configuration and instantiation system to use OmegaConf for flexible, Hydra-style configuration and dynamic object construction. It also improves project scaffolding, updates the dataset/model/trainer interfaces for better modularity, and cleans up legacy code. The most important changes are outlined below.

**Configuration System Overhaul:**

* Switched configuration loading and dumping to use OmegaConf, supporting both Pydantic and DictConfig objects, and updated all related utilities and templates accordingly. (`pyproject.toml`, `trainite/config/base.py`, `trainite/cli/init.py`, [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R13) [[2]](diffhunk://#diff-73352dfb97a0b9afb3ad346b22af2fdb5df6e72c73da6aa0df96d85c0a4c0cddL38-R50) [[3]](diffhunk://#diff-1e20c19b0f95a48591a0f35c025ef834b85ff6c4deaecb675a711f8499f1bf40L41-L79) [[4]](diffhunk://#diff-1e20c19b0f95a48591a0f35c025ef834b85ff6c4deaecb675a711f8499f1bf40L93-L102)
* Updated config classes to include a `_target_` field with a default value, enabling dynamic instantiation of datasets and models via OmegaConf/Hydra patterns. (`trainite/config/dataset.py`, `trainite/config/model.py`, [[1]](diffhunk://#diff-cc556a921ff67b5954e6a97505d0e02344c19dcd16e3cd057473afd7355b07eeL3-R10) [[2]](diffhunk://#diff-6e7f91c6e3bffd7c468fc89b41bb721e64cb819cf2e9a19e7f899213fc232a43L1-R7)

**Project Initialization and Templates:**

* Refactored project scaffolding to generate a more modular file structure (separate `models/`, `dataset/`, `utils.py`), and updated template logic to match the new config and import patterns. (`trainite/cli/init.py`, [[1]](diffhunk://#diff-1e20c19b0f95a48591a0f35c025ef834b85ff6c4deaecb675a711f8499f1bf40R102-R115) [[2]](diffhunk://#diff-1e20c19b0f95a48591a0f35c025ef834b85ff6c4deaecb675a711f8499f1bf40L159-R133) [[3]](diffhunk://#diff-1e20c19b0f95a48591a0f35c025ef834b85ff6c4deaecb675a711f8499f1bf40L181-R174) [[4]](diffhunk://#diff-1e20c19b0f95a48591a0f35c025ef834b85ff6c4deaecb675a711f8499f1bf40L277-R261)

**Model/Dataset/Trainer Interface Updates:**

* Refactored model and dataset builder functions to accept plain arguments (matching config fields) and arbitrary keyword arguments, decoupling them from Pydantic config classes and enabling direct instantiation from OmegaConf configs. (`trainite/models/transformer.py`, `trainite/datasets/string_reverse.py`, [[1]](diffhunk://#diff-23286b30be7f6477ea835453fdb3dcc78f5fe95323349abe4f0c6ebf17fe04e8L66-R88) [[2]](diffhunk://#diff-23286b30be7f6477ea835453fdb3dcc78f5fe95323349abe4f0c6ebf17fe04e8L96-R117) [[3]](diffhunk://#diff-e0dac8bceead0c2b475a6440a83d3cb3f966218558961b9832d2eeb75dbb4250L31-R60)
* Updated the `PreTrainer` class to use a generic `instantiate` utility for model and dataset creation, and to accept additional training parameters directly. (`trainite/trainers/pretrainer.py`, [[1]](diffhunk://#diff-8f40d97481e59dd82f8477965646f5d9517f80b514ebf0d45e17346dae39c1caL1-R46) [[2]](diffhunk://#diff-8f40d97481e59dd82f8477965646f5d9517f80b514ebf0d45e17346dae39c1caL93-R100)

**General Cleanup:**

* Removed unnecessary `from __future__ import annotations` statements and unused imports across multiple files. (`trainite/cli/init.py`, `trainite/config/registry.py`, `trainite/config/trainer.py`, `trainite/datasets/string_reverse.py`, `trainite/models/transformer.py`, [[1]](diffhunk://#diff-1e20c19b0f95a48591a0f35c025ef834b85ff6c4deaecb675a711f8499f1bf40L1-L2) [[2]](diffhunk://#diff-f1e4797054e117a0de73729f410d0b1781a627ce2ff13b9d02d2cbf0e5fc1a4bL1-L2) [[3]](diffhunk://#diff-b17caca894810164f85d88750aff84f166bdcafaeebe3c796ca521980e1d18bdL1-L2) [[4]](diffhunk://#diff-e0dac8bceead0c2b475a6440a83d3cb3f966218558961b9832d2eeb75dbb4250L1-L2) [[5]](diffhunk://#diff-23286b30be7f6477ea835453fdb3dcc78f5fe95323349abe4f0c6ebf17fe04e8L1-L2)

These changes modernize the configuration and instantiation process, making the codebase more extensible and compatible with Hydra/OmegaConf best practices.

### DIFF
--- a/trainite/cli/init.py
+++ b/trainite/cli/init.py
@@ -4,7 +4,7 @@ import textwrap
 from pathlib import Path
 from typing import Iterable, Sequence
 
-from trainite.config import default_config, dump_config
+from trainite.config import OutputConfig, ProjectConfig, dump_config
 from trainite.config.registry import (
     REGISTRY,
     get_dataset_spec,
@@ -34,41 +34,6 @@ def _render_template(path: Path, replacements: Iterable[tuple[str, str]] = ()) -
 
 def _render_class_source(cls: type) -> str:
     return textwrap.dedent(inspect.getsource(cls)).strip()
-
-
-def _build_config_template() -> str:
-    lines = [
-        "from pathlib import Path",
-        "from typing import Any",
-        "",
-        "import yaml",
-        "from omegaconf import OmegaConf, DictConfig, ListConfig",
-        "from pydantic import BaseModel, Field",
-        "",
-        "",
-        "def load_yaml(path: str | Path) -> dict:",
-        "    config_path = Path(path)",
-        "    data = yaml.safe_load(config_path.read_text()) or {}",
-        "    if not isinstance(data, dict):",
-        '        raise ValueError("Config file must contain a mapping")',
-        "    return data",
-        "",
-        "",
-        "def dump_yaml(data: dict, path: str | Path) -> None:",
-        "    Path(path).write_text(yaml.safe_dump(data, sort_keys=False))",
-        "",
-        "",
-        "def load_config(path: str | Path) -> DictConfig | ListConfig:",
-        "    return OmegaConf.load(path)",
-        "",
-        "",
-        "def dump_config(config: DictConfig, path: str | Path) -> None:",
-        "    data = OmegaConf.to_container(config, resolve=True)",
-        "    Path(path).write_text(yaml.safe_dump(data, sort_keys=False))",
-        "",
-        "",
-    ]
-    return "\n".join(lines).strip() + "\n"
 
 
 def _prompt_text(prompt: str, default: str) -> str:
@@ -105,13 +70,12 @@ def _write_file(path: Path, content: str, force: bool) -> None:
 
 
 def _build_templates(
-    project_name: str, model_name: str, dataset_name: str, trainer_name: str
+    model_name: str, dataset_name: str, trainer_name: str
 ) -> dict[str, str]:
     model_spec = get_model_spec(model_name)
     dataset_spec = get_dataset_spec(dataset_name)
     trainer_spec = get_trainer_spec(trainer_name)
     return {
-        "config.py": _build_config_template(),
         f"models/{model_spec.name}.py": _render_template(
             model_spec.implementation_path,
             model_spec.template_replacements,
@@ -126,14 +90,15 @@ def _build_templates(
         ),
         "utils.py": _render_template(
             PROJECT_ROOT / "trainite/utils.py",
-            [],
+            [("trainite.config", "config")],
         ),
+        "config.py": _render_template(PROJECT_ROOT / "trainite/config/base.py", []),
         "main.py": _render_template(
             PROJECT_ROOT / "trainite/main.py",
             [
                 (
-                    "from trainite.config import default_config, load_config",
-                    "from config import default_config, load_config",
+                    "trainite.config",
+                    "config",
                 ),
                 (
                     "from trainite.trainers import PreTrainer",
@@ -167,6 +132,9 @@ def build_parser() -> argparse.ArgumentParser:
     init_parser.add_argument("--output-root", help="Output root for generated config")
     init_parser.add_argument("--run-name", help="Run name for generated config")
     init_parser.add_argument(
+        "--trainer", choices=TRAINER_CHOICES, help="Starter trainer template to use"
+    )
+    init_parser.add_argument(
         "--force",
         action="store_true",
         help="Overwrite existing starter files",
@@ -183,54 +151,57 @@ def build_parser() -> argparse.ArgumentParser:
 
 
 def init_project(args: argparse.Namespace) -> None:
-    defaults = default_config()
-    if args.project_dir:
-        project_dir_input = args.project_dir
-    elif args.yes:
-        project_dir_input = "my-lm-experiment"
-    else:
-        project_dir_input = _prompt_text("Project directory", "my-lm-experiment")
-    project_dir = _project_directory(project_dir_input, args.force)
-
     if args.yes:
-        model_name = args.model or defaults.model_name
-        dataset_name = args.dataset or defaults.dataset_name
-        output_root = args.output_root or defaults.output.root
+        project_dir = args.project_dir or "my-cool-experiment"
+        model_name = args.model or MODEL_CHOICES[0]
+        dataset_name = args.dataset or DATASET_CHOICES[0]
+        trainer_name = args.trainer or TRAINER_CHOICES[0]
+        output_root = args.output_root or "outputs"
         run_name = args.run_name or f"{model_name}__{dataset_name}"
     else:
+        project_dir = args.project_dir or _prompt_text(
+            "Project directory", "my-cool-experiment"
+        )
         model_name = args.model or _prompt_choice(
-            "Model", MODEL_CHOICES, defaults.model_name
+            "Model", MODEL_CHOICES, MODEL_CHOICES[0]
         )
         dataset_name = args.dataset or _prompt_choice(
-            "Dataset", DATASET_CHOICES, defaults.dataset_name
+            "Dataset", DATASET_CHOICES, DATASET_CHOICES[0]
         )
-        output_root = args.output_root or _prompt_text(
-            "Output directory", defaults.output.root
+        trainer_name = args.trainer or _prompt_choice(
+            "Trainer", TRAINER_CHOICES, TRAINER_CHOICES[0]
         )
+        output_root = args.output_root or _prompt_text("Output directory", "outputs")
         run_name = args.run_name or _prompt_text(
             "Run name", f"{model_name}__{dataset_name}"
         )
 
-    starter_config = default_config()
-    starter_config.output.root = output_root
-    starter_config.output.run_name = run_name
+    project_dir = _project_directory(project_dir, args.force)
 
-    project_name = project_dir.name
+    output_config = OutputConfig(root=output_root, run_name=run_name)
 
     # Build templates for the starter project
-    templates = _build_templates(
-        project_name, model_name, dataset_name, defaults.trainer_name
-    )
+    templates = _build_templates(model_name, dataset_name, trainer_name)
 
     model_spec = get_model_spec(model_name)
     dataset_spec = get_dataset_spec(dataset_name)
+    trainer_spec = get_trainer_spec(trainer_name)
 
     # Update config to point to the correct builder functions for the model and dataset
-    starter_config.model.target = (
-        f"models.{model_spec.name}.{model_spec.builder_symbol}"
-    )
-    starter_config.dataset.target = (
+    model_component = model_spec.config_cls()
+    dataset_component = dataset_spec.config_cls()
+    trainer_component = trainer_spec.config_cls()
+
+    model_component.target = f"models.{model_spec.name}.{model_spec.builder_symbol}"
+    dataset_component.target = (
         f"dataset.{dataset_spec.name}.{dataset_spec.builder_symbol}"
+    )
+
+    starter_config = ProjectConfig(
+        model=model_component,
+        dataset=dataset_component,
+        trainer=trainer_component,
+        output=output_config,
     )
 
     dump_config(starter_config, project_dir / "config.yaml")
--- a/trainite/cli/init.py
+++ b/trainite/cli/init.py
@@ -132,18 +132,20 @@ def _project_directory(raw_project_dir: str, force: bool) -> Path:
 def _write_file(path: Path, content: str, force: bool) -> None:
     if path.exists() and not force:
         raise FileExistsError(f"{path} already exists; pass --force to overwrite it")
+    if path.parent:
+        path.parent.mkdir(parents=True, exist_ok=True)
     path.write_text(content)
 
 
 def _build_templates(
-    model_name: str, dataset_name: str, trainer_name: str
+    project_name: str, model_name: str, dataset_name: str, trainer_name: str
 ) -> dict[str, str]:
     model_spec = get_model_spec(model_name)
     dataset_spec = get_dataset_spec(dataset_name)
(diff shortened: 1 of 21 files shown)


aaishwarymishra:
https://github.com/pytorch-ignite/trainite/blob/0b55ef8ca09ca6aa6a2eb3e96f11100c563292a5/trainite/cli/init.py#L39

currently we are using this hardcoded string for generating `config.py` because the `trainite/config/base.py` also have definition of the projectconfig. Should we make separate file for this, or use original `base.py`.

trainite/cli/init.py
```diff
@@ -274,12 +251,14 @@ def init_project(args: argparse.Namespace) -> None:
         )
 
     starter_config = default_config()
```
aaishwarymishra: Also we are using default_config() to basically generate the default pydantic project class, I am not sure if this is the way we should handle default values, thoughts?

aaishwarymishra:
Also currently we can set the `output_dir`  either through the yaml or the cli, but we can only set it inside the project directory, should we make it global?

trainite/config/dataset.py
```diff
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class StringReverseDatasetConfig(BaseModel):
+    target: str = Field(
+        default="trainite.datasets.string_reverse.build_string_reverse_dataloaders",
```
TahaZahid05, line 8: Should point to local files generated instead?

trainite/config/model.py
```diff
@@ -1,9 +1,10 @@
-from __future__ import annotations
-
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class TransformerModelConfig(BaseModel):
+    target: str = Field(
+        default="trainite.models.transformer.build_transformer_model", alias="_target_"
```
TahaZahid05, line 8: same here

trainite/cli/init.py
```diff
@@ -38,45 +36,15 @@ def _render_class_source(cls: type) -> str:
...
         "",
         "import yaml",
+        "from omegaconf import OmegaConf, DictConfig",
         "from pydantic import BaseModel, Field",
         "",
-        model_source,
-        "",
-        dataset_source,
-        "",
-        trainer_source,
```
TahaZahid05, line 62: If we dont copy the Pydantic classes, the user won't get autocomplete? We can use Pydantic and OmegConf together. using `OmegaConf.merge` to check the YAML against the Pydantic class giving us the advantage of IDE as well.

trainite/models/transformer.py
TahaZahid05: can remove import as not needed anymore

trainite/cli/init.py
```diff
@@ -90,16 +58,15 @@ def _build_config_template(
         "    Path(path).write_text(yaml.safe_dump(data, sort_keys=False))",
         "",
         "",
-        "def load_config(path: str | Path) -> ProjectConfig:",
-        "    return ProjectConfig.model_validate(load_yaml(path))",
+        "def load_config(path: str | Path) -> Any:",
```
TahaZahid05: Typehint should stay `ProjectConfig`?

trainite/config/base.py
```diff
@@ -35,16 +34,20 @@ def load_yaml(path: str | Path) -> dict:
...
+    if hasattr(config, "model_dump"):
+        data = config.model_dump(by_alias=True)
+    else:
+        data = OmegaConf.to_container(config, resolve=True)
+    dump_yaml(data, path)
 
 
-def load_config(path: str | Path) -> ProjectConfig:
-    return ProjectConfig.model_validate(load_yaml(path))
+def load_config(path: str | Path) -> Any:
```
TahaZahid05: Same here, type hint should stay `ProjectConfig`

TahaZahid05:
> https://github.com/pytorch-ignite/trainite/blob/0b55ef8ca09ca6aa6a2eb3e96f11100c563292a5/trainite/cli/init.py#L39
> 
> currently we are using this hardcoded string for generating `config.py` because the `trainite/config/base.py` also have definition of the projectconfig. Should we make separate file for this, or use original `base.py`.

I think we should use the original `inspect.getsource` method but then add the logic on top to merge such that we can leverage both `OmegaConf` and `Pydantic`

TahaZahid05:
> Also currently we can set the `output_dir` either through the yaml or the cli, but we can only set it inside the project directory, should we make it global?

Yeah, we should make it global.

I think default should still be `output/` inside the project folder so a new user doesn't get confused. Your thoughts?

aaishwarymishra:
@TahaZahid05 Should I merge it? you can pull this change and update your pr, as I have changed a few things on how cli works , there might be conflicts.

I think we can talk about type hinting with pydantic classes on next meeting or when victor gets time.

Currently we don't generate readme and project.toml should I try to add it too or make seperate pr.

trainite/config/registry.py
```diff
@@ -30,37 +31,58 @@ class DatasetSpec(ComponentSpec):
...
 
 
+MODEL_SPECS = {
+    "transformer": ModelSpec(
+        name="transformer",
+        implementation_path=Path("trainite/models/transformer.py"),
+        config_cls=TransformerModelConfig,
+        implementation_symbol="TransformerModel",
+        builder_symbol="build_transformer_model",
+        template_replacements=[],
```
TahaZahid05, line 41: Shouldn't this have the import statements that you removed in `init.py`?

trainite/config/registry.py
```diff
@@ -32,37 +31,58 @@ class DatasetSpec(ComponentSpec):
...
+}
+
+DATASET_SPECS = {
+    "string-reverse": DatasetSpec(
+        name="string-reverse",
+        implementation_path=Path("trainite/datasets/string_reverse.py"),
+        config_cls=StringReverseDatasetConfig,
+        implementation_symbol="StringReverseDataset",
+        builder_symbol="build_string_reverse_dataloaders",
+        template_replacements=[],
```
TahaZahid05, line 52: Same here

trainite/config/registry.py
```diff
@@ -32,37 +31,58 @@ class DatasetSpec(ComponentSpec):
...
+TRAINER_SPECS = {
+    "pretrainer": TrainerSpec(
+        name="pretrainer",
+        implementation_path=Path("trainite/trainers/pretrainer.py"),
+        config_cls=PreTrainerConfig,
+        implementation_symbol="PreTrainer",
+        builder_symbol="PreTrainer",
+        template_replacements=[
+            (
+                "from trainite.config import ProjectConfig, dump_config",
```
TahaZahid05: ProjectConfig is no more being imported in `pretrainer.py`. The template string is inaccurate so the replacement isn't taking place.

trainite/cli/init.py
```diff
@@ -132,79 +99,50 @@ def _project_directory(raw_project_dir: str, force: bool) -> Path:
...
+            PROJECT_ROOT / "trainite/utils.py",
+            [],
         ),
         "main.py": _render_template(
-            PROJECT_ROOT / "main.py",
+            PROJECT_ROOT / "trainite/main.py",
             [
                 (
                     "from trainite.config import default_config, load_config",
                     "from config import default_config, load_config",
```
TahaZahid05: `default_config` is not being imported anymore.

TahaZahid05:
@aaishwarymishra I have requested some more changes. We can talk about pydantic related changes in the meeting. Regarding `pyproject.toml` and `README.MD`, we can create a seperate PR for them once we are done with this. I would recommend that you generate an experiment with trainite, and see if the user files are all as expected and no dependencies on trainite. Currently the generate files are not standalone.

trainite/config/registry.py
aaishwarymishra: They are redundant now I checked.

trainite/config/registry.py
aaishwarymishra: Ok there I am confused , like should I just replace trainer.config to config?

trainite/cli/init.py
aaishwarymishra: here too, what do you think?

trainite/config/registry.py
aaishwarymishra: @TahaZahid05

TahaZahid05: https://github.com/pytorch-ignite/trainite/blob/b2757a073392fafa7d3cf05374e32360e1c40058/trainite/trainers/pretrainer.py#L14

I was suggesting:
```python
"from trainite.config import dump_config"
```
currently your cli isn't catching "from trainite.config import ProjectConfig, dump_config" as you have changed it in `pretrainer.py` as shown above.

trainite/cli/init.py
TahaZahid05: just change the string to match what you have written in `main.py`, so it can catch the template string:

https://github.com/pytorch-ignite/trainite/blob/b2757a073392fafa7d3cf05374e32360e1c40058/trainite/main.py#L7

trainite/cli/init.py
TahaZahid05: Yeah, currently this is quite static (will not work for other models except Transformer). We would need to handle this by changing `ProjectConfig` pydantic class to be generic (`Any` rather than transformer specific). Then, move `starter_config` initailization after we have asked the user the model and dataset they want to use, and pass those args to `default_config()` so that it dynamically makes the yaml.

aaishwarymishra: I have implemented it pls check.

aaishwarymishra:
ok, so I have introduced a pydantic project structure, I haven't added optimizer conf yet I will add it with the scheduler I guess. Fixed most of the problems, generated code doesn't depend on trainite , global output folder support. I think if its look good, we can merge it as base and you can try to add your code and dataset over it.

trainite/trainers/pretrainer.py
```diff
@@ -20,7 +19,7 @@
 class PreTrainer:
     def __init__(
         self,
-        config: DictConfig,
+        config: ProjectConfig,
         device: str | torch.device = "cpu",
         learning_rate: float = 1e-3,
```
TahaZahid05: Let's set this to None as default so it doesn't override the config's learning rate.

trainite/config/base.py
```diff
@@ -2,28 +2,31 @@
...
-    model: Any = Field(default_factory=TransformerModelConfig)
-    dataset: Any = Field(default_factory=StringReverseDatasetConfig)
-    trainer: Any = Field(default_factory=PreTrainerConfig)
-    output: OutputConfig = Field(default_factory=OutputConfig)
+    model: ComponentConfig
+    dataset: ComponentConfig
+    trainer: TrainerConfig
+    output: OutputConfig
     seed: int = 42
+    lr: float = 1e-3
```
TahaZahid05: We don't need `lr` here. we already have `learning_rate` in `PreTrainerConfig`. This was creating two different learning rate parameters in the generated yaml.

trainite/trainers/pretrainer.py
```diff
@@ -42,7 +41,9 @@ def __init__(
         self.model = model or instantiate(config.model)
         self.model.to(self.device)
         self.loss_fn = nn.CrossEntropyLoss()
-        self.optimizer = torch.optim.AdamW(self.model.parameters(), lr=learning_rate)
+        self.optimizer = torch.optim.AdamW(
+            self.model.parameters(), lr=learning_rate or config.lr
```
TahaZahid05: use `config.trainer.learning_rate` as lr would be there now once u remove it from `ProjectConfig`.

TahaZahid05 (commented):
Left my final comments. Everything else looks good!

trainite/trainers/pretrainer.py
aaishwarymishra: Fixed it.

trainite/trainers/pretrainer.py
```diff
@@ -21,41 +20,45 @@ class PreTrainer:
     def __init__(
         self,
         config: ProjectConfig,
+        device: str | torch.device | None = None,
+        learning_rate: float | None = None,
+        epochs: int | None = None,
+        log_every_steps: None = None,
```
```suggestion
        log_every_steps: int | None = None,
```

trainite/config/trainer.py
```diff
@@ -1,17 +1,9 @@
...
-from pydantic import BaseModel
 
-
-class PreTrainerConfig(BaseModel):
-    device: str = "cpu"
+class PreTrainerConfig(TrainerConfig):
     type: str = "pre"
     epochs: int = 3
     learning_rate: float = 1e-3
     log_every_steps: int = 10
```
TahaZahid05, line 8:
```suggestion
```

Already the default values as `TrainerConfig` so no need to write again.

TahaZahid05 (approved):

TahaZahid05:
@aaishwarymishra you can merge after committing the suggestions.

trainite/config/trainer.py
aaishwarymishra: I think its fine , we want to overwrite it anyways so let it be there so we wont forget to change it in future to more suitable value

PR #3, opened 2026-05-12
StringReverseDataset character version and variable length support

TahaZahid05:
## Overview
This PR updates the `StringReverseDataset` to use characters instead of integers and adds support for variable-length sequences.

### Dataset & Model
- **Alphabet-based data**: The dataset now generates sequences using a configurable alphabet (default `a-z`).
- **Variable Lengths**: Added `min_seq_len`, `max_seq_len`, and `fixed_length` settings.
- **Padding**: Reserved Index `0` for padding. Added masking in `CrossEntropyLoss` and Accuracy so padding tokens are ignored during training and evaluation.
- **Decoding**: Added a `.decode()` method to convert IDs back into characters.

### Framework & CLI
- **Modular Config**: Updated the `PreTrainer` to take `config.dataset` directly.
- **Pydantic Validation**: Updated type-hints in `ProjectConfig` so YAML configs load correctly into objects.
- **CLI Initialization**: Updated the `trainite init` logic to handle the new function calls when scaffolding projects.
- **Example Update**: Updated the `examples/string-reverse` project to use these changes.

### DIFF
--- a/examples/string-reverse/config.py
+++ b/examples/string-reverse/config.py
@@ -4,7 +4,25 @@ from typing import Any
 import torch
 import yaml
 from omegaconf import OmegaConf
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+
+ALPHABET_PRESETS = {
+    "@alpha": "abcdefghijklmnopqrstuvwxyz",
+    "@digits": "0123456789",
+    "@alphanumeric": "abcdefghijklmnopqrstuvwxyz0123456789",
+}
+
+
+def resolve_alphabet(alphabet: str) -> str:
+    return ALPHABET_PRESETS.get(alphabet, alphabet)
+
+
+def _component_value(component: BaseModel, name: str) -> Any:
+    value = getattr(component, name, None)
+    if value is not None:
+        return value
+    return component.model_extra.get(name) if component.model_extra else None
 
 
 class OutputConfig(BaseModel):
@@ -31,6 +49,29 @@ class ProjectConfig(BaseModel):
     seed: int = 42
     device: str = "cuda" if torch.cuda.is_available() else "cpu"
 
+    @model_validator(mode="after")
+    def validate_vocab_size(self) -> "ProjectConfig":
+        dataset_vocab_size = _component_value(self.dataset, "vocab_size")
+        alphabet = _component_value(self.dataset, "alphabet")
+
+        if dataset_vocab_size is not None and alphabet is not None:
+            expected_vocab_size = len(resolve_alphabet(str(alphabet)))
+            if int(dataset_vocab_size) != expected_vocab_size:
+                raise ValueError(
+                    "dataset.vocab_size is derived from dataset.alphabet; "
+                    f"expected {expected_vocab_size}, got {dataset_vocab_size}"
+                )
+
+        model_vocab_size = _component_value(self.model, "vocab_size")
+        if model_vocab_size is not None and dataset_vocab_size is not None:
+            if int(model_vocab_size) < int(dataset_vocab_size):
+                raise ValueError(
+                    "model.vocab_size must be greater than or equal to "
+                    "dataset.vocab_size"
+                )
+
+        return self
+
 
 def load_yaml(path: str | Path) -> dict:
     config_path = Path(path)
@@ -51,4 +92,5 @@ def dump_config(config: ProjectConfig, path: str | Path) -> None:
 
 def load_config(path: str | Path) -> ProjectConfig:
     raw_conf = OmegaConf.load(path)
-    return ProjectConfig.model_validate(raw_conf)
+    data = OmegaConf.to_container(raw_conf, resolve=True)
+    return ProjectConfig.model_validate(data)
--- a/examples/string-reverse/config.py
+++ b/examples/string-reverse/config.py
@@ -1,54 +1,35 @@
-from __future__ import annotations
-
 from pathlib import Path
+from typing import Any
 
+import torch
 import yaml
-from pydantic import BaseModel, Field
-
-class TransformerModelConfig(BaseModel):
-    vocab_size: int = 32
-    hidden_size: int = 64
-    num_layers: int = 2
-    num_heads: int = 2
-    feedforward_dim: int = 128
-    dropout: float = 0.1
-    max_seq_len: int = 128
-
-class StringReverseDatasetConfig(BaseModel):
-    alphabet: str = "abcdefghijklmnopqrstuvwxyz"
-    vocab_size: int = 26
-    train_size: int = 256
-    val_size: int = 64
-    batch_size: int = 32
-    min_seq_len: int = 1
-    max_seq_len: int = 16
-    fixed_length: bool = True
-    num_workers: int = 2
-    seed: int = 7
-
-class PreTrainerConfig(BaseModel):
-    device: str = "cpu"
-    type: str = "pre"
-    epochs: int = 3
-    learning_rate: float = 1e-3
-    log_every_steps: int = 10
-    grad_clip_norm: float | None = None
+from omegaconf import OmegaConf
+from pydantic import BaseModel, ConfigDict, Field
 
 
 class OutputConfig(BaseModel):
-    root: str = "output"
-    run_name: str = "transformer__string-reverse"
+    root: str
+    run_name: str
+
+
+class ComponentConfig(BaseModel):
+    model_config = ConfigDict(extra="allow")
+    target: str = Field(alias="_target_")
+
+
+class TrainerConfig(BaseModel):
+    model_config = ConfigDict(extra="allow")
+    learning_rate: float = 1e-3
+    log_every_steps: int = 10
 
 
 class ProjectConfig(BaseModel):
-    model_name: str = "transformer"
-    dataset_name: str = "string-reverse"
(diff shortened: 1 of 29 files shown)


TahaZahid05:
Currently, `base.py` uses specific type-hints:

https://github.com/pytorch-ignite/trainite/blob/8e27f1f583b4386fcf7d931d93f0d8161bced814/trainite/config/base.py#L21-L23

We need a more generic validation strategy later on so it remains model agnostic.

examples/string-reverse/config.py
```diff
@@ -15,12 +15,15 @@ class TransformerModelConfig(BaseModel):
     max_seq_len: int = 128
 
 class StringReverseDatasetConfig(BaseModel):
-    vocab_size: int = 32
+    alphabet: str = "abcdefghijklmnopqrstuvwxyz"
```
aaishwarymishra: I am not sure we should take the alphabets or not, we can just take type, whether alpha or alpha numeric.

examples/string-reverse/config.py
```diff
@@ -15,12 +15,15 @@ class TransformerModelConfig(BaseModel):
...
+    vocab_size: int = 26
```
aaishwarymishra: Also whats the use of vocab_size here? Its understandable in the model but not sure in the dataset.

examples/string-reverse/dataset.py
```diff
@@ -2,56 +2,115 @@
...
+    def decode(self, ids: torch.Tensor) -> str:
+        return "".join([self.id_to_char[idx.item()] for idx in ids if idx.item() != 0])
+
+
+def collate_fn(batch: list[dict[str, torch.Tensor]]) -> dict[str, torch.Tensor]:
+    input_ids = [item["input_ids"] for item in batch]
+    labels = [item["labels"] for item in batch]
+    
+    padded_input_ids = pad_sequence(input_ids, batch_first=True, padding_value=0)
+    padded_labels = pad_sequence(labels, batch_first=True, padding_value=0)
```
aaishwarymishra: I think we should take default padding of -100 , instead of the 0, it would also made us stop using vocab_size+1

examples/string-reverse/config.py
TahaZahid05: Good point. Maybe we should give both options? The strings can provide flexibility incase they want to test against a specific subset of characters but maybe utitlies like: `alphabet: "@digits"` or `alphabet: "@lowercase"` for shortcuts to help with the issue you have pointed out. What do you think?

examples/string-reverse/dataset.py
TahaZahid05: True, default padding should be -100. I'll fix that. But curious, how would this lead to stop using vocab_size + 1? We would still need that becaues `nn.Embedding` require a positive index for padding meaning we need the 0th index for it.

examples/string-reverse/config.py
aaishwarymishra: I am not sure I understand what you said.

examples/string-reverse/dataset.py
aaishwarymishra: when we are using 0 as padding, 0 will not be considered an index of a character , with 0 as padding vocab index stars from 1, and the vocab size would be vocab+1

examples/string-reverse/config.py
TahaZahid05: I am saying user should have option of both, adding custom alphabets and some presets. like:

```python
# Standard presets
alphabet: "@alpha"    # a-z
alphabet: "@digits"   # 0-9

# Or custom research sets
alphabet: "abc!@#"   
```

examples/string-reverse/dataset.py
TahaZahid05: Can you show through example? I think it doesn't matter whether you use -100 or 0 as label, the input still needs a number to act as a placeholder for the padding in a batch.

Since that placeholder number can't be one of our 26 letters, we always end up needing 27 slots (26 letters + 1 placeholder).

aaishwarymishra: here https://github.com/pytorch-ignite/trainite/pull/3/changes#diff-e0dac8bceead0c2b475a6440a83d3cb3f966218558961b9832d2eeb75dbb4250R21-R23

So basically you were using the 0 as padding your starting index for mapping is 0+1, if you had say 6 strings the last index will be 5+1, as last index is 6 so your vocab_size is 6.
Embedding layer expects the indexes or ids to start from 0, so if your vocab_size is 6 it expects valid indexes to be in between 0-5, here we have reserved 0 for padding so valid index for us will be 1-6, thus we need to set vocab_size+1.

TahaZahid05: I understand that.. How does using `-100` solve this? You still require an index for padding which results in vocab_size+1?

aaishwarymishra: Pytorch layers ignore -100 by default as they treat it as padding . So you do not  need to explicitly do vocab_size + 1.

TahaZahid05: `nn.Embedding` doesn't ignore it, which means we still have to give it a space. `nn.CrossEntropyLoss` ignores it thats why I changed to -100. I think you are confusing Loss and Embedding.

aaishwarymishra: Yeah my bad.

examples/string-reverse/config.py
aaishwarymishra: sounds good

examples/string-reverse/dataset.py
aaishwarymishra: For labels the padding value should be -100, if taking 0 we have to set it explicitly in loss function in the trainer

trainite/models/transformer.py
```diff
@@ -69,7 +69,7 @@ def __init__(
         **kwargs,
     ) -> None:
         super().__init__()
-        self.embedding = nn.Embedding(vocab_size, hidden_size)
+        self.embedding = nn.Embedding(vocab_size + 1, hidden_size, padding_idx=0)
```
aaishwarymishra, line 72: As default, should we give user the option to change padding or enable or disable by default?  I am not sure

trainite/datasets/string_reverse.py
```diff
@@ -1,60 +1,126 @@
...
+        self,
+        size: int,
+        min_seq_len: int,
+        max_seq_len: int,
+        seed: int,
+        fixed_length: bool = True,
+        alphabet: str = "abcdefghijklmnopqrstuvwxyz",
+    ) -> None:
+        self.alphabet = ALPHABET_PRESETS.get(alphabet, alphabet)
+        vocab_size = len(self.alphabet)
```
aaishwarymishra, line 24: Here vocab_size in dynamically set but in model we have to set it manually, I am not sure if its correct pattern.

aaishwarymishra:
I just noticed at root there is `config.py`, its not getting used anywhere I may have forgot to remove it.

trainite/datasets/string_reverse.py
TahaZahid05: Well, that is because dataset's vocab_size is derived from the `alphabet` string. Whereas, model's vocab_size is the capacity of the model's embedding as to the maximum string length it can process. It is not something we can add dynamically based on the dataset as models can be trained on certain lengths of strings and then tested on even larger strings to check for OOD. Does that make sense?

aaishwarymishra: what about say we define vocab size in datasets, say 64. During the string generation phase it takes first 64 characters from the alphabet like `alphabet[:64]`. I am not sure if its a good idea or not though.

trainite/models/transformer.py
TahaZahid05: I don't think we should give option to disable padding as that is a critical requirement for batching in model training.. There is no use of disabling padidngs?

Also, for changing padding. I don't see how that can affect training or be useful in experiments. Your thoughts?

aaishwarymishra: I think user can add it himself if he wants so its fine ig

trainite/datasets/string_reverse.py
TahaZahid05: That can be quite ambiguous because what if the user wants certain parts of the `alphabet` in their `vocab_size` defined rather than just the first 64. I think the most flexible approach is that they use the shorthands or specify the characters they want to be used and then we calculate length dynamically based on their input.

aaishwarymishra: My main concern is for the model, user wont know exact vocab size needed for training the model if its getting calculated dynamically, either he has to set the value high, or dynamically get the vocb_size

TahaZahid05: Good point. Maybe, we can write `vocab_size` for dataset into the yaml but explicitly mark it as a derived value by adding a comment, and also some validation to see if it matches the desired length. This  way the user can see the dataset's `vocab_size` so they can set model's `vocab_size` accordingly. 

Thoughts?

TahaZahid05:
@aaishwarymishra as a temporary approach, I have made the model's `vocab_size` default value point to dataset's `vocab_size`. This approach still has cons of the user manually changing it if they decide to modify `alphabet`. We may need to discuss a better approach together in tomorrow's meeting. I would suggest we can merge this for now and deal with the design issue later after we talk about it.

aaishwarymishra:
Okay you can merge it whenever it's ready 👍

PR #15, opened 2026-05-19
Add optimizer config and learning rate scheduler

aaishwarymishra:
This pull request introduces a more flexible and configurable optimizer and learning rate scheduler setup, improving how training hyperparameters are managed and making the training pipeline more modular. The main changes are the introduction of an `OptimizerConfig` class, switching to instantiating the optimizer and scheduler from configuration, and updating the minimum required Python version.

**Configuration and Optimizer Improvements:**

* Added a new `OptimizerConfig` class to `trainite/config/base.py`, allowing the optimizer type and learning rate to be specified in the configuration. The optimizer is now included as a field in `ProjectConfig`, making it easier to customize and extend optimizer settings.
* Updated the trainer in `trainite/trainers/pretrainer.py` to instantiate the optimizer from the new `config.optimizer` rather than hardcoding `AdamW`, supporting more flexible optimizer selection.

**Learning Rate Scheduler:**

* Introduced a two-phase learning rate scheduler using `SequentialLR` and `LinearLR` in the trainer, with a warmup and cooldown phase based on the total number of training iterations. The scheduler is stepped every iteration for smoother learning rate transitions. [[1]](diffhunk://#diff-8f40d97481e59dd82f8477965646f5d9517f80b514ebf0d45e17346dae39c1caR11) [[2]](diffhunk://#diff-8f40d97481e59dd82f8477965646f5d9517f80b514ebf0d45e17346dae39c1caL43-R69) [[3]](diffhunk://#diff-8f40d97481e59dd82f8477965646f5d9517f80b514ebf0d45e17346dae39c1caL152-R178)

**Python Version Requirement:**

* Lowered the minimum required Python version from 3.13 to 3.10 in `pyproject.toml` for broader compatibility.Instantiate optimizer from project config (params are passed) instead of using a hardcoded AdamW. Add a SequentialLR schedule with a short linear warmup followed by linear decay, stepped every iteration. Relax Python requirement to >=3.10 and update the lockfile. Remove trailing whitespace in string-reverse dataset files.


trainite/trainers/pretrainer.py
```diff
@@ -149,10 +167,15 @@ def _log_loss(engine):
             _log_loss,
         )
 
-        # 2. Run evaluations
+        # 2. Step LR scheduler every iteration
+        self.engine.add_event_handler(
+            Events.ITERATION_STARTED, lambda _: self.scheduler.step()
+        )
```
aaishwarymishra, line 165: Wasn't able to use ignites `LrScheduler` wrapper as `SequentialLr` doesnt implement `.get_lr()`

I dont know we can remove the warmup stage if we want to use it not sure, what you suggest should we keep warmup or not?

trainite/trainers/pretrainer.py
TahaZahid05: we can remove this?

TahaZahid05:
https://github.com/pytorch-ignite/trainite/blob/dbc4e380b50ecb11f09db9f0b9d921e598ee8d80/trainite/config/trainer.py#L7

We should remove this as well.

trainite/trainers/pretrainer.py
TahaZahid05: We can use ignite's [`create_lr_scheduler_with_warmup`](https://docs.pytorch.org/ignite/generated/ignite.handlers.param_scheduler.create_lr_scheduler_with_warmup.html#ignite.handlers.param_scheduler.create_lr_scheduler_with_warmup)

trainite/trainers/pretrainer.py
```diff
@@ -149,10 +167,15 @@ def _log_loss(engine):
             _log_loss,
         )
 
-        # 2. Run evaluations
+        # 2. Step LR scheduler every iteration
+        self.engine.add_event_handler(
+            Events.ITERATION_STARTED, lambda _: self.scheduler.step()
```
TahaZahid05: Shouldn't we use `ITERATION_COMPLETED`? we should call `scheduler.setp()` after `optimizer.step()`

trainite/trainers/pretrainer.py
```diff
@@ -40,15 +41,32 @@ def __init__(
...
 
+        total_iters = len(self.train_loader) * self.epochs
+        self.scheduler = SequentialLR(
+            self.optimizer,
+            schedulers=[
+                LinearLR(
+                    self.optimizer,
+                    start_factor=0.1,
+                    end_factor=1.0,
+                    total_iters=int(total_iters * 0.1),
```
TahaZahid05: This may crash on very small runs where `total_iters=0`. wrap it in `max(1,int(total_iters * 0.1))`

trainite/trainers/pretrainer.py
```diff
@@ -149,10 +149,27 @@ def _log_loss(engine):
             _log_loss,
         )
 
-        # 2. Run evaluations
+        # 2. Step LR scheduler every iteration
+        warmup_iters = int(0.1 * self.total_iters)
```
TahaZahid05: https://github.com/pytorch/ignite/blob/9debb725b1d547897bea053f5c883713b7cb1ee1/ignite/handlers/param_scheduler.py#L1142

Requires `max(2,..`) still

trainite/trainers/pretrainer.py
```diff
@@ -149,10 +149,27 @@ def _log_loss(engine):
...
+            total_iters=self.total_iters - warmup_iters,
+        )
+        self.scheduler = create_lr_scheduler_with_warmup(
+            linear_decay,
+            warmup_start_value=0.0,
+            warmup_end_value=self.lr,
+            warmup_duration=warmup_iters,
+        )
+
+        self.engine.add_event_handler(Events.ITERATION_STARTED, self.scheduler)
```
TahaZahid05: We should step after optimizer so use `Events.ITERATION_COMPLETED`

TahaZahid05 (approved):

PR #16, opened 2026-05-19
Decouple Tokenization & Dynamic Device Assignment

TahaZahid05:
## Summary
This PR refactors Trainite's architecture to decouple tokenization from hardcoded dataset implementations, transforming it into a first-class configurable component within the registry system. Additionally, it transitions the default training device from hardcoded `"cuda"` to `"auto"` (resolved dynamically at runtime) to support GPU-less systems out-of-the-box without requiring heavyweight `torch` checks at CLI project-generation time.

---

### 1. Tokenizer as a Component Config
* **New Tokenizer Module**: Added `trainite/tokenizers/` housing a standard `CharTokenizer` supporting custom alphabets, alphabet presets (`@alpha`, `@digits`, `@alphanumeric`), and standard methods (`encode`, `decode`, `vocab_size`, `pad_token_id`).
* **Pydantic Config Schema**: Defined `CharTokenizerConfig` using the `ComponentConfig` factory mapping. Added `tokenizer: ComponentConfig` as a mandatory field on `ProjectConfig`.
* **Component Registry**: Registered `char` tokenizer under `TOKENIZER_SPECS` inside `registry.py` for CLI templating.

### 2. Architectural Decoupling & Injection
* **Model Configuration**: Removed hardcoded `vocab_size` from `TransformerModelConfig` and its YAML mapping.
* **Dataset Configuration**: Removed hardcoded `alphabet` from `StringReverseDatasetConfig`. 
* **Orchestration**: Updated `main.py` and the `PreTrainer` engine to:
  1. Dynamically instantiate the tokenizer from the YAML config.
  2. Compute `vocab_size` from the instantiated tokenizer and feed it into the model constructor at runtime.
  3. Inject the `tokenizer` instance directly into the dataset builder, which uses `tokenizer.vocab_size` for synthetic sequence generation.

### 3. CLI Templating & Example Regeneration
* Extended `trainite init` CLI tool to support a `--tokenizer` flag.
* Added generation logic to create a `tokenizers/` subdirectory in starter projects.
* Wrote dynamic import-rewriting replacements for the template runner, including a bugfix that ensures custom/alternative trainers (like `RLTrainer`) correctly swap out the hardcoded `PreTrainer` class in generated `main.py` entrypoints.
* Regenerated the `examples/string-reverse/` directory to fully showcase the new tokenizer setup.

### 4. Dynamic Device Resolution
* Set the default value of the top-level `device` config in `ProjectConfig` to `"auto"` instead of `"cuda"`.
* Updated `PreTrainer` to resolve `"auto"` at runtime based on `torch.cuda.is_available()`. This guarantees that generated templates compile and run seamlessly on CPU-only machines while retaining GPU acceleration when present, all without adding a heavy `torch` runtime import dependency to the lightweight CLI init step.

---


aaishwarymishra:
ok, like I have 2 main questions:
1. Do we need to have tokenizers? the goal of the trainite is to be a lightweight training tool kit.
2. For string reverse the tokenizer is super simple, we could simply have a tokenizer class or method , which can be returned by say dataset builder method.

The thing is I am not sure to have a complete logic for tokenizer generation.

Also in prototype should we write tests? because I have not written tests for many of the features I added, might have to add them now, or when we start to work on main branch. Not sure.

aaishwarymishra:
Also what about ci/cd etc?

TahaZahid05:
> ok, like I have 2 main questions:
> 
> 1. Do we need to have tokenizers? the goal of the trainite is to be a lightweight training tool kit.
> 2. For string reverse the tokenizer is super simple, we could simply have a tokenizer class or method , which can be returned by say dataset builder method.
> 
> The thing is I am not sure to have a complete logic for tokenizer generation.
> 
> Also in prototype should we write tests? because I have not written tests for many of the features I added, might have to add them now, or when we start to work on main branch. Not sure.

1) Well, for actual real research, people use a pretrained tokenizer all the time from Hugging Face. By separating the tokenizer, they can edit the YAML more easily to change the tokeniser:

```yaml
tokenizer:
  _target_: "transformers.AutoTokenizer.from_pretrained"
  pretrained_model_name_or_path: "meta-llama/Meta-Llama-3-8B"
```

2) I agree that in a string reverse task, its overhead seems more expansive than its advantage but this is a toy example. I suppose when used for more complex tasks, the tokeniser's overhead would be justified.

It's not necessary to write tests right now but i guess it still helps to write them as you make new features, as it would be harder to make all the test cases in one pass at the end, but it is fine either way.

TahaZahid05:
> Also what about ci/cd etc?

We can talk about this in the next meeting with Victor.

aaishwarymishra:
> 1. Well, for actual real research, people use a pretrained tokenizer all the time from Hugging Face. By separating the tokenizer, they can edit the YAML more easily to change the tokeniser:

I get it what you trying to say, but I think you should consult with victor about it. But our uses cases are simple, so it seems overkill.

trainite/config/base.py
```diff
@@ -34,7 +34,7 @@ class ProjectConfig(BaseModel):
     trainer: TrainerConfig
     output: OutputConfig
     seed: int = 42
-    device: str = "cuda"
+    device: str = "auto"
```
aaishwarymishra, line 37: I dont think auto is supported by the torch.

trainite/cli/init.py
```diff
@@ -190,13 +193,17 @@ def _build_templates(
...
+                ),
                 (
                     "from trainite.trainers import PreTrainer",
                     f"from trainer import {trainer_spec.implementation_symbol}",
                 ),
                 (
-                    "trainer = PreTrainer(config)",
-                    f"trainer = {trainer_spec.implementation_symbol}(config)",
+                    "    trainer = PreTrainer(",
+                    f"    trainer = {trainer_spec.implementation_symbol}(",
```
aaishwarymishra: why there is these extra space?

trainite/config/base.py
TahaZahid05: Using "auto" here because at start we are not sure about user's device.  in `pretrainer.py`, we use it to then assign device to user based on the device available. This solves the inital usage of trainite to be free of torch dependency while also removing errors for the user due to device as it can be easily fixed using this.

I think the previous method of hardcoding cuda to be default device is error prone.

trainite/cli/init.py
```diff
@@ -190,13 +193,17 @@ def _build_templates(
...
+                ),
                 (
                     "from trainite.trainers import PreTrainer",
                     f"from trainer import {trainer_spec.implementation_symbol}",
                 ),
                 (
-                    "trainer = PreTrainer(config)",
-                    f"trainer = {trainer_spec.implementation_symbol}(config)",
+                    "trainer = PreTrainer(",
+                    f"trainer = {trainer_spec.implementation_symbol}(",
```
aaishwarymishra, line 206: why this change?

TahaZahid05:
Closed in favor of #18

PR #18, opened 2026-05-21
Switch to Decoder-Only Transformers, Decouple Tokenization & Dynamic Device Assignment

TahaZahid05:
# Description
This PR transitions the baseline transformer template to a **true Decoder-Only architecture** and automates runtime device, vocabulary-size, and sequence-level tokenization configurations.

### Key Enhancements

1. **Decoder-Only Causal Attention**
   * Modified `TransformerBlock` in `models/transformer.py` to dynamically generate and apply a square subsequent causal attention mask (`torch.nn.Transformer.generate_square_subsequent_mask`). This prevents the model from looking ahead during autoregressive training.

2. **Autoregressive Sequence Alignment & Masking**
   * Configured the `string-reverse` dataset to format training sequences as:
     `[BOS] + original_string + [EOS] + reversed_string + [EOS]`
   * Aligned inputs and targets by shifting `target_labels` by 1 (`full_seq[1:]`) relative to `input_ids` (`full_seq[:-1]`).
   * Applied standard prompt masking by setting loss targets for the prompt (`[BOS] + original_string + [EOS]`) to `-100`, forcing the model to only compute loss and learn on predicting the reversed string tokens.

3. **Modular Tokenizer Class**
   * Extracted a standalone `CharTokenizer` class inside `datasets/string_reverse.py` handling all token maps, special character allocations (`pad=0`, `bos=1`, `eos=2`), encoding, and decoding independently.

4. **Auto Device & Vocab-Size Detection**
   * Replaced the hardcoded `"cuda"` training device in `ProjectConfig` with `"auto"`, which dynamically resolves to GPU if available.
   * Completely removed the static `vocab_size` from `model.py` configuration. The model's vocabulary size is now dynamically extracted from the dataset at runtime, with Pydantic support to handle manual overrides when defined in `config.yaml`.

### DIFF
--- a/trainite/datasets/string_reverse.py
+++ b/trainite/datasets/string_reverse.py
@@ -17,6 +17,7 @@ class CharTokenizer:
     ID 0 is reserved as the padding token.
     ID 1 is reserved as the BOS token.
     ID 2 is reserved as the EOS token.
+    ID 3 is reserved as the UNK token.
     """
 
     def __init__(self, alphabet: str = "abcdefghijklmnopqrstuvwxyz") -> None:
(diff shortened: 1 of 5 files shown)


trainite/trainers/pretrainer.py
```diff
@@ -63,6 +61,26 @@ def __init__(
...
+
+        if configured_vocab_size is not None:
+            if self.vocab_size is not None and configured_vocab_size < self.vocab_size:
+                raise ValueError(
+                    f"Configured model vocab_size ({configured_vocab_size}) is smaller than "
+                    f"the dataset vocabulary size ({self.vocab_size}). "
+                    f"Please increase model vocab_size or remove it from config.yaml "
+                    f"to let it resolve automatically."
+                )
+            self.vocab_size = configured_vocab_size
```
aaishwarymishra, line 77: okay, we are giving error  when val_size in model config is smaller , but over ridding if it bigger, I think its not a good approach, either we override in  both cases and give user the error, or we give error on smaller size but keep the defined config size if its bigger.

trainite/datasets/string_reverse.py
```diff
@@ -10,6 +10,43 @@
...
+        return [self.char_to_id[c] for c in text]
+
+    def decode(self, ids: list[int] | torch.Tensor) -> str:
+        """Convert a list of token IDs back to a string, skipping padding and special tokens."""
+        if isinstance(ids, torch.Tensor):
+            ids = ids.tolist()
+        special_token_ids = {self.pad_token_id, self.bos_token_id, self.eos_token_id}
+        return "".join(
+            self.id_to_char[i] for i in ids if i not in special_token_ids and i in self.id_to_char
+        )
```
aaishwarymishra: decoder will fail if we try to say decode labels which has been padded with -100, and I think we should add a fallback if a character is not in vocab we mark it as a special character we identify it as special token UNK or something like that.

Also currently decode ignores special character, I think decode should also return special characters like bos,eos , unk etc. We can add a parameter whether to ignore them or not.

trainite/trainers/pretrainer.py
TahaZahid05: There can be cases where a model can have a bigger `vocab_size` than the dataset such as fine tuning on a small, specific task. That is why I allowed a larger vocab_size if passed, and gave an error on smaller vocab_size since that doesn't make sense. Your thoughts?

trainite/datasets/string_reverse.py
TahaZahid05: done

PR #19, opened 2026-05-22
Integrate `FBResearchLogger`.

TahaZahid05:
## Description
This PR integrates PyTorch-Ignite's `FBResearchLogger` into the training pipeline to provide clean, structured console and file progress logging out of the box.
### Key Changes
* **Two-Stage Logging**:
  * **Constructor Stage**: Initialized `self.logger = setup_logger("trainer")` early to safely handle pre-run warning logs.
  * **Late Upgrade Stage**: Upgraded the logger with `reset=True` and a file path inside `_attach_handlers` once the run directory is lazily generated.
* **`FBResearchLogger` Attachment**: Attached to the training engine to track running loss and optimizer learning rates.
* **OOP Encapsulation**: Unified all internal logging statements to use the encapsulated `self.logger` attribute.


trainite/trainers/pretrainer.py
```diff
@@ -177,9 +178,24 @@ def _attach_metrics(self) -> None:
...
+            name="Train",
+            every=self.log_every_steps,
+            optimizer=self.optimizer,
+            output_transform=lambda output: {"loss": output["loss"].item()},
+        )
+
         # 1. Log training loss every N steps
         def _log_loss(engine):
-            logger.info(
+            self.logger.info(
```
aaishwarymishra: why use this ? like I am not sure the fb_logger `every` and we give it `self.log_every_steps` right does it do the same task ?

trainite/trainers/pretrainer.py
```diff
@@ -177,9 +178,24 @@ def _attach_metrics(self) -> None:
...
+            filepath=str(self.run_dir / "output.log") if self.run_dir else None,
+            reset=True,
+        )
+        self.train_fb_logger = FBResearchLogger(logger=self.logger, show_output=True)
+        self.train_fb_logger.attach(
+            self.engine,
+            name="Train",
+            every=self.log_every_steps,
+            optimizer=self.optimizer,
+            output_transform=lambda output: {"loss": output["loss"].item()},
```
aaishwarymishra, line 194: I think we should also log other metrics, like currently we are only calculating loss and accuracy, are there some other important metric that is eligible for this task? we should add it in _attach_metrics and log it too.

aaishwarymishra:
can you also check like `main.py`, there is some logger setting in that too.

trainite/trainers/pretrainer.py
TahaZahid05: apologies, you are right. there is no reason to log this again.

TahaZahid05:
@aaishwarymishra I checked `main.py`. No changes needed there, they are fine.

aaishwarymishra (approved):

PR #22, opened 2026-05-23
Added exact accuracy

TahaZahid05:
Currently, the `PreTrainer` computes token-level accuracy (categorical accuracy). This PR introduces the **Exact Accuracy Metric** (sequence-level exact match), assigning a score of 1 if the predicted sequence exactly matches the ground truth (excluding ignored target/padding tokens) and 0 otherwise.

## Changes
- **`trainite/trainers/pretrainer.py`**:
  - Replaced the token-level `_flatten_accuracy` function with `_exact_accuracy_transform`.
  - The new transform compares predicted tokens and targets at all non-ignored/non-masked positions to check for exact sequence match.
  - Returns binary long tensors `(y_pred, y)` where `y_pred` is `1` if all masked tokens match and `0` otherwise, while `y` is all `1`s.
  - Attached standard `Accuracy` metric to evaluator engine using `_exact_accuracy_transform` under the new metric key `"accuracy"`.
  - Updated all references to `"token_accuracy"` (TensorBoard logger, checkpoint `score_function`, and logger evaluations print) to `"accuracy"`.

### DIFF
--- a/trainite/trainers/pretrainer.py
+++ b/trainite/trainers/pretrainer.py
@@ -176,9 +176,9 @@ class PreTrainer:
         val_accuracy = Accuracy(output_transform=self._exact_accuracy_transform)
 
         train_loss.attach(self.train_evaluator, "loss")
-        train_accuracy.attach(self.train_evaluator, "accuracy")
+        train_accuracy.attach(self.train_evaluator, "exact_accuracy")
         val_loss.attach(self.val_evaluator, "loss")
-        val_accuracy.attach(self.val_evaluator, "accuracy")
+        val_accuracy.attach(self.val_evaluator, "exact_accuracy")
 
         self.metrics = {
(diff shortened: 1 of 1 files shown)


trainite/trainers/pretrainer.py
```diff
@@ -218,7 +227,7 @@ def _attach_handlers(self) -> None:
         to_save = {"model": self.model, "optimizer": self.optimizer}
 
         def score_function(engine):
-            val_acc = engine.state.metrics["token_accuracy"]
+            val_acc = engine.state.metrics["accuracy"]
```
aaishwarymishra: I think we should log it as exact_accuracy, to not get confused with normal accuracy

aaishwarymishra: wherever accuracy is used

PR #30, opened 2026-05-31
Add Encoder-Decoder model support

TahaZahid05:
# PR Description:

This Pull Request introduces the **Encoder-Decoder (Seq2Seq)** transformer architecture option to the Trainite CLI toolbox. It fully decouples encoder and decoder hyperparameters, implements robust sequence-to-sequence template routing, and establishes a highly thorough, symmetrical testing suite across all datasets, models, and trainers.

---

### 1. Symmetrical & Decoupled Model Architecture
* Implemented the encoder-decoder transformer in `trainite/models/encoder_decoder.py` with:
  * Unified `MultiHeadAttention` supporting causal self-attention, non-causal self-attention, and cross-attention.
  * Separate `EncoderBlock` and `DecoderBlock` layer stacks.
  * Sine/Cosine positional encodings tailored individually to encoder and decoder context lengths.
* Decoupled hyperparameters cleanly in `EncoderDecoderModelConfig` (`num_encoder_layers` vs `num_decoder_layers`, `encoder_max_seq_len` vs `decoder_max_seq_len`).

### 2. Dynamic Template Routing (CLI)
* Renamed CLI choice `transformer` to `decoder` for symmetric clarity.
* Integrated the `--model encoder-decoder` option in the CLI (`trainite/cli/init.py`).
* Implemented a dynamic routing system that delivers highly tailored, focused files for either model choice without embedding nested `if/else` checks or branching logic into user-generated boilerplates.

### 3. Specialized Seq2Seq Dataset & Trainer Templates
* **Dataset (`trainite/datasets/string_reverse_seq2seq.py`)**: Yields separate source and target tensors with bespoke padding collation (handling target pads with `ignore_index=-100` and inputs with `0`).
* **Trainer (`trainite/trainers/pretrainer_seq2seq.py`)**: Implements clean pretraining and evaluation loops specifically for sequence-to-sequence forward passes.

### 4. Fully Aligned and Comprehensive Testing
* Expanded the unit test suite to be 100% symmetrical between decoder and encoder-decoder paths.
* **Model Tests (`tests/models/encoder_decoder_test.py`)**: Added positional encoding dimension checks, spec-based model instantiation, and deterministic/random dropout evaluations.
* **Dataset Tests (`tests/datasets/string_reverse_seq2seq_test.py`)**: Added checks for special tokens, character preset variations, sequence lengths, and exception handling.
* **Trainer Tests (`tests/trainers/pretrainer_seq2seq_test.py`)**: Integrated trainer lifecycle tests verifying loss flattening, exact accuracy conversions, device selections, and last-checkpoint fallbacks.

### DIFF
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -190,4 +190,3 @@ def test_generated_encoder_decoder_project_is_runnable() -> None:
             )
         except subprocess.TimeoutExpired:
             pytest.fail("Generated project timed out")
-
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -12,7 +12,8 @@ from trainite.config import get_dataset_spec, get_model_spec
 @pytest.mark.parametrize(
     "model,dataset,trainer",
     [
-        ("transformer", "string-reverse", "pretrainer"),
+        ("decoder", "string-reverse", "pretrainer"),
+        ("encoder-decoder", "string-reverse", "pretrainer"),
     ],
 )
 def test_init_generates_valid_project(model: str, dataset: str, trainer: str) -> None:
@@ -81,7 +82,7 @@ def test_generated_string_reversal_project_is_runnable() -> None:
                 "init",
                 "--yes",
                 "--model",
-                "transformer",
+                "decoder",
                 "--dataset",
                 "string-reverse",
                 "--trainer",
@@ -126,3 +127,67 @@ def test_generated_string_reversal_project_is_runnable() -> None:
             )
         except subprocess.TimeoutExpired:
             pytest.fail("Generated project timed out")
+
+
+def test_generated_encoder_decoder_project_is_runnable() -> None:
+    """
+    Test that the generated encoder-decoder project can actually be run for a few steps.
+    """
+    with tempfile.TemporaryDirectory() as temp_dir:
+        project_dir = Path(temp_dir) / "runnable-enc-dec-project"
+
+        # 1. Generate project
+        subprocess.run(
+            [
+                sys.executable,
+                "-m",
+                "trainite.cli",
+                "init",
+                "--yes",
+                "--model",
+                "encoder-decoder",
+                "--dataset",
+                "string-reverse",
+                "--trainer",
+                "pretrainer",
+                str(project_dir),
+            ],
+            check=True,
+        )
+
+        # 2. Modify config.yaml to run for only 1 step/epoch to keep test fast
+        config_path = project_dir / "config.yaml"
+        import yaml
+
+        with open(config_path, "r") as f:
+            config = yaml.safe_load(f)
+
+        config["trainer"]["epochs"] = 1
+        config["trainer"]["log_every_steps"] = 1
+        config["model"]["num_encoder_layers"] = 1
+        config["model"]["num_decoder_layers"] = 1
+        config["model"]["hidden_size"] = 16
+        config["model"]["feedforward_dim"] = 32
+        config["model"]["num_heads"] = 2
+        config["data"]["train"]["dataset"]["size"] = 16
+
+        with open(config_path, "w") as f:
+            yaml.safe_dump(config, f)
+
+        # 3. Run the generated main.py
+        try:
+            subprocess.run(
+                [sys.executable, "main.py", "config.yaml"],
+                cwd=project_dir,
+                check=True,
+                timeout=60,
+                capture_output=True,
+                text=True,
+            )
+        except subprocess.CalledProcessError as e:
+            pytest.fail(
+                f"Generated project failed to run:\nSTDOUT: {e.stdout}\nSTDERR: {e.stderr}"
+            )
+        except subprocess.TimeoutExpired:
+            pytest.fail("Generated project timed out")
+
--- a/trainite/cli/init.py
+++ b/trainite/cli/init.py
@@ -169,11 +169,19 @@ def _build_templates(
 
     if model_name == "encoder-decoder":
         if dataset_name == "string-reverse":
-            dataset_implementation_path = Path("trainite/datasets/string_reverse_seq2seq.py")
-            dataset_readme_path = Path("trainite/templates/components/datasets/string_reverse_seq2seq.md")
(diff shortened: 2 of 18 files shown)


tests/cli_test.py
```diff
@@ -126,3 +127,66 @@ def test_generated_string_reversal_project_is_runnable() -> None:
...
+            ],
+            check=True,
+        )
+
+        # 2. Modify config.yaml to run for only 1 step/epoch to keep test fast
+        config_path = project_dir / "config.yaml"
+        import yaml
+
+        with open(config_path, "r") as f:
+            config = yaml.safe_load(f)
```
aaishwarymishra, line 163: I think we have helper method for yaml dumping and loading in config.py

tests/cli_test.py
```diff
@@ -126,3 +127,66 @@ def test_generated_string_reversal_project_is_runnable() -> None:
...
+        config["trainer"]["log_every_steps"] = 1
+        config["model"]["num_encoder_layers"] = 1
+        config["model"]["num_decoder_layers"] = 1
+        config["model"]["hidden_size"] = 16
+        config["model"]["feedforward_dim"] = 32
+        config["model"]["num_heads"] = 2
+        config["data"]["train"]["dataset"]["size"] = 16
+
+        with open(config_path, "w") as f:
+            yaml.safe_dump(config, f)
```
aaishwarymishra, line 175: same

trainite/cli/init.py
```diff
@@ -160,6 +160,29 @@ def _build_templates(
...
+                "trainite/templates/components/datasets/string_reverse_seq2seq.md"
+            )
+        if trainer_name == "pretrainer":
+            trainer_implementation_path = Path(
+                "trainite/trainers/pretrainer_seq2seq.py"
+            )
+            trainer_readme_path = Path(
+                "trainite/templates/components/trainers/pretrainer_seq2seq.md"
+            )
+
```
aaishwarymishra, line 185: Dont understand use of it, I thought dataset, trainer and model would be independent you might have to consult with victor, on how modular should the generated code be, can user select different components or based on the use case.

trainite/models/encoder_decoder.py
```diff
@@ -0,0 +1,306 @@
...
+                dropout_p=self.dropout_p if self.training else 0.0,
+            )
+        else:
+            context = nn.functional.scaled_dot_product_attention(
+                q,
+                k,
+                v,
+                is_causal=is_causal,
+                dropout_p=self.dropout_p if self.training else 0.0,
+            )
```
aaishwarymishra, line 95: This if else is super weird looking also you didnt even create context in the first if statement.

trainite/models/encoder_decoder.py
```diff
@@ -0,0 +1,306 @@
...
+        super().__init__()
+        self.embed_dim = embed_dim
+        self.num_heads = num_heads
+        self.head_dim = embed_dim // num_heads
+        if not self.num_heads * self.head_dim == self.embed_dim:
+            raise ValueError("embed_dim must be divisible by num_heads.")
+
+        self.q_proj = nn.Linear(embed_dim, embed_dim, bias=False)
+        self.k_proj = nn.Linear(embed_dim, embed_dim, bias=False)
+        self.v_proj = nn.Linear(embed_dim, embed_dim, bias=False)
```
aaishwarymishra, line 41: we can have a single  multi dim linear layer for key and value as it would be both either x in case of self and y in case of cross.

trainite/models/encoder_decoder.py
```diff
@@ -0,0 +1,306 @@
...
+        **kwargs,
+    ) -> None:
+        super().__init__()
+        self.embedding = nn.Embedding(vocab_size, hidden_size, padding_idx=0)
+        self.encoder_pos_encoding = PositionalEncoding(
+            hidden_size, encoder_max_seq_len, dropout
+        )
+        self.decoder_pos_encoding = PositionalEncoding(
+            hidden_size, decoder_max_seq_len, dropout
+        )
```
aaishwarymishra, line 207: we could use single layer for this

trainite/models/encoder_decoder.py
```diff
@@ -0,0 +1,306 @@
...
+        else:
+            dec_padding_mask = None
+
+        # Cross attention padding mask (B_dec, 1, 1, S_enc) to mask out padded tokens in encoder memory
+        if (encoder_input_ids == self.embedding.padding_idx).any():
+            cross_padding_mask = (
+                encoder_input_ids != self.embedding.padding_idx
+            ).reshape(B_dec, 1, 1, S_enc)
+        else:
+            cross_padding_mask = None
```
aaishwarymishra, line 269: its same as encoder self padding?

aaishwarymishra:
Also like I dont think we need new datasets for this model old reverse-string should work and the old trainer should work too.

trainite/models/encoder_decoder.py
TahaZahid05: I don't understand, context is in both conditionals?

trainite/models/encoder_decoder.py
TahaZahid05: I wanted to keep them decoupled incase someone wants to try and modify either one of encoder or decoder in the setup.

TahaZahid05:
> Also like I dont think we need new datasets for this model old reverse-string should work and the old trainer should work too.

I would discuss this with victor as well, but having the same dataset and pretrainer for both cases can lead to if/else statements inside the file which can be more confusing for the end-user. That is why I opted for creating a seperate file to keep it clean and only the code required for the model they chose.

trainite/models/encoder_decoder.py
aaishwarymishra: I think the model should be simple enough, if your wants separate encoder , decoder position encoder they can add it themselves

trainite/models/encoder_decoder.py
aaishwarymishra: my bad I got confused by nested if else

PR #32, opened 2026-06-01
Add dataset config: explicit splits or auto-split

aaishwarymishra:
Implement two data configuration styles (explicit per-split configs or a single dataset with train/val ratios) with Pydantic validation in DataConfig. Update trainer to support automatic splitting (random_split), adapt configs/tests/examples (string_reverse) to use per_seq_size, update docs/templates, add tests, and fix related bugs (e.g. positional encoding d_model check).


trainite/datasets/string_reverse.py
```diff
@@ -59,13 +64,6 @@ def decode(
         skip_special_tokens: bool = True,
         ignore_index: int = -100,
     ) -> str:
-        """Convert a list of token IDs back to a string.
```
TahaZahid05, line 62: why remove this?

trainite/cli/init.py
```diff
@@ -280,6 +280,25 @@ def build_parser() -> argparse.ArgumentParser:
     return parser
 
 
+def _update_targets(config: Any, old_prefix: str, new_prefix: str) -> None:
```
TahaZahid05: Just keep union of the different configs rather than Any. you can remove import as well.

trainite/config/base.py
```diff
@@ -40,10 +46,52 @@ class SplitConfig(BaseModel):
...
+
+    @model_validator(mode="after")
+    def validate_options(self) -> "DataConfig":
+        option1_fields = {"train", "val", "test"}
+        option2_fields = {"dataset", "train_ratio", "val_ratio", "dataloader"}
+
+        present_option1 = {f for f in option1_fields if getattr(self, f) is not None}
+        present_option2 = {f for f in option2_fields if getattr(self, f) is not None}
+
+        if present_option1 and present_option2:
```
TahaZahid05, line 68: Why have sub-conditional errors when you are giving error in both conditions. Just have one value error that tells the user that they have to use either one of the configurations?

trainite/datasets/string_reverse.py
```diff
@@ -113,55 +118,81 @@ def __init__(
...
 
-            input_ids = full_seq[:-1]
-            target_labels = full_seq[1:].clone()
+                self.inputs.append(input_ids)
+                self.labels.append(target_labels)
 
-            prompt_len = len(seq) + 2
-            target_labels[: prompt_len - 1] = -100
+        # Shuffle so variable-length samples are evenly distributed across batches
+        final_shuffle_gen = torch.Generator().manual_seed(seed + 1)
```
TahaZahid05: I think remove `+ 1` for uniformity with other places where seed is used.

trainite/trainers/pretrainer.py
```diff
@@ -36,43 +43,37 @@ def __init__(
...
+        self.set_loaders(config, train_loader, val_loader, test_loader)
+        if self.train_loader is None:
+            raise ValueError(
+                "At least a training loader must be provided either directly or via config"
+            )
+        train_dataset = self.train_loader.dataset
+        # Handle Subset (from random_split) which wraps the original dataset
+        if isinstance(train_dataset, torch.utils.data.Subset):
+            train_dataset = train_dataset.dataset
+        self.vocab_size: int = getattr(train_dataset, "vocab_size")
```
TahaZahid05: Why remove `None`?

tests/trainers/pretrainer_test.py
```diff
@@ -283,3 +298,137 @@ def test_pretrainer_test_without_val(project_config, temp_run_dir):
...
+    assert isinstance(trainer.train_loader.sampler, torch.utils.data.RandomSampler)
+
+
+def test_trainer_option_2(tmp_path):
+    config = ProjectConfig(
+        model=cc(
+            "trainite.models.transformer.build_transformer_model",
+            vocab_size=100,
+            d_model=32,
+            nhead=2,
```
TahaZahid05: These arguments don't exist. the test is running with default values. Use `hidden_size` and `num_heads`

trainite/config/base.py
```diff
@@ -40,10 +46,52 @@ class SplitConfig(BaseModel):
...
     val: SplitConfig | None = None
     test: SplitConfig | None = None
 
+    # Option 2: Automatic splitting
+    dataset: ComponentConfig | None = None
+    dataloader: DataLoaderConfig | None = None
+    train_ratio: float | None = None
+    val_ratio: float | None = None
+
+    @model_validator(mode="after")
```
TahaZahid05, line 60: Why not validate the ratios here rather than in pretrainer?

trainite/trainers/pretrainer.py
aaishwarymishra: because for our specific case we need vocab_size for model creation. so instead of taking a None as fallback we just give error here saving time.

trainite/config/base.py
aaishwarymishra: we can have dataset, train_split etc, if we dont have this sub-conditional check then we would just give a super generic message and user will get confused

aaishwarymishra:
@TahaZahid05  check now :)

trainite/datasets/string_reverse.py
```diff
@@ -59,7 +64,7 @@ def decode(
         skip_special_tokens: bool = True,
         ignore_index: int = -100,
     ) -> str:
-        """Convert a list of token IDs back to a string.
+        """Convert a list of token IDs back to a string.Expand commentComment on line L62
```
TahaZahid05: Kindly fix.

trainite/trainers/pretrainer.py
TahaZahid05: User can have `vocab_size` defined in config but not as an attribute. Your implementation will raise an error rather than setting to vocab_size to the value in config. The previous code checked whether vocab_size was undefined in both attribute and config before giving the error.

PR #38, opened 2026-06-03
WIP: added interactive pudb run command

vfdev-5:
`trainite i_run config.yaml`


vfdev-5:
Adding interactive pudb to trainite still makes sense but in another PR

PR #40, opened 2026-06-04
Added optional hyperparameter sweep support using Optuna.

TahaZahid05:
## What's New

* **Optuna Integration**: Core sweep runner logic in `sweep.py` and `sweep_utils.py` supporting `grid`, `random`, and `tpe` search strategies.
* **CLI Updates**: Added a `--sweep` flag to `trainite init` to generate a boilerplate `sweep.yaml` configuration, runner files, and add `optuna` dependency to `pyproject.toml`.
* **Validation**: Added `SweepConfig` and `ParameterRange` Pydantic schemas to validate target parameters and their search spaces.
* **Tests**: Added new tests for configuration validation, parameter overrides, and CLI execution in `tests/config/base_test.py` and `tests/cli_test.py`.


aaishwarymishra:
@TahaZahid05 can you tell me how generated code will look like?

trainite/config/sweep.py
```diff
@@ -0,0 +1,164 @@
...
+        if self.low >= self.high:
+            raise ValueError(f"low ({self.low}) must be less than high ({self.high})")
+        if self.sample == "log" and self.low <= 0:
+            raise ValueError("low must be positive when sample='log'")
+        if self.step is not None and self.step <= 0:
+            raise ValueError("step must be positive")
+        return self
+
+
+class PruningConfig(BaseModel):
```
aaishwarymishra, line 27: This class is super verbose, I think instead of us managing and handling every case, we could simply use a default pruning technique instead. If user wants he can edit the code.

trainite/config/sweep.py
```diff
@@ -0,0 +1,164 @@
+from pathlib import Path
+from typing import Any, Literal
+
+import yaml
+from pydantic import BaseModel, ConfigDict, model_validator
+
+
+class ParameterRange(BaseModel):
+    model_config = ConfigDict(validate_assignment=True, extra="forbid")
+    type: Literal["int", "float"]
```
aaishwarymishra, line 10: categorical can also be present

trainite/config/sweep.py
```diff
@@ -0,0 +1,164 @@
...
+        "gp",
+        "nsgaii",
+        "qmc",
+    ] = "grid"
+    direction: Literal["maximize", "minimize"] = "maximize"
+    metric: str = "exact_accuracy"
+    n_trials: int | None = None
+    storage: str | None = None
+    pruning: PruningConfig = PruningConfig()
+    parameters: dict[str, list[Any] | ParameterRange]
```
aaishwarymishra, line 142: A dict config will be bad here I think because for every parameter, there will be an extra nested arguments.

I think we should use a list here, with __target__  here being consistent with other configs.

trainite/templates/project/README.md
```diff
@@ -13,6 +13,7 @@ Trainite is a toolbox for training language models with PyTorch-Ignite. This pro
 - `trainer.py`: Defines the training and evaluation logic. You can override `train_step` or `eval_step` here.
 - `config.py`: Contains Pydantic models for configuration validation. If you add new parameters to `config.yaml`, update the models here.
 - `utils.py`: Shared utilities for configuration and logging.
+{{sweep_project_structure}}
```
aaishwarymishra, line 16: Sweeping is a completly extra feature, so I think we should generate a separate readme for it entierely.

aaishwarymishra:
Also about the metric_name, and other parameters, they will be unique to use case to use case right? hmm, I am not sure how would we generate them for different setup.

I think you should consult victor about it, if we should even support sweeping or not as it was more of an prototype to check if it should be added or not.

TahaZahid05:
@aaishwarymishra should `SweepConfig` also have only a default strategy or like three stratigies like grid, random, and tpe only and the rest can be defined by user if needed? Asking since you pointed out the verbosity in `PruningConfig` and this class also struggles with it a bit ig. thoughts? Also, I am thinking of keeping median as default for `SweepConfig`. does that sound fine?

aaishwarymishra:
> @aaishwarymishra should `SweepConfig` also have only a default strategy or like three stratigies like grid, random, and tpe only and the rest can be defined by user if needed? Asking since you pointed out the verbosity in `PruningConfig` and this class also struggles with it a bit ig. thoughts? Also, I am thinking of keeping median as default for `SweepConfig`. does that sound fine?

yes, I think the main functionality the sweeper should provide is the easy way to run multiple experiments with different values. we should not try to handle every possible combination of parameter like strategy, we provide default, user can change it however they want. This keeps code and yaml clean.

PR #41, opened 2026-06-06
Watching test output

aaishwarymishra:
Pr for inference support.

### DIFF
--- a/trainite/cli/init.py
+++ b/trainite/cli/init.py
@@ -378,11 +378,13 @@ def init_project(args: argparse.Namespace) -> None:
         collate_target = model_spec.collate_fn_target
 
         def _inject_collate(config):
-            if hasattr(config, "dataloader") and config.dataloader is not None:
-                config.dataloader.collate_fn = ComponentConfig(_target_=collate_target)
+            dataloader = getattr(config, "dataloader", None)
+            if dataloader is not None:
+                dataloader.collate_fn = ComponentConfig(_target_=collate_target)
             for split in ["train", "val", "test"]:
-                if hasattr(config, split) and getattr(config, split) is not None:
-                    _inject_collate(getattr(config, split))
+                split_config = getattr(config, split, None)
+                if split_config is not None:
+                    _inject_collate(split_config)
 
         _inject_collate(data_config)
 
--- a/trainite/cli/init.py
+++ b/trainite/cli/init.py
@@ -373,23 +373,36 @@ def init_project(args: argparse.Namespace) -> None:
     data_config = dataset_spec.config_cls()
     trainer_component = trainer_spec.config_cls()
 
-    # Update targets to point to the local project structure
-    _update_targets(
-        model_component,
-        f"trainite.models.{model_spec.name}",
-        f"models.{model_spec.name}",
-    )
-    _update_targets(
-        data_config,
-        f"trainite.datasets.{dataset_spec.name}",
-        f"datasets.{dataset_spec.name}",
-    )
+    # Inject model collator into data config dataloaders
+    if model_spec.collate_fn_target:
+        collate_target = model_spec.collate_fn_target
+
+        def _inject_collate(config):
+            if hasattr(config, "dataloader") and config.dataloader is not None:
+                config.dataloader.collate_fn = ComponentConfig(_target_=collate_target)
+            for split in ["train", "val", "test"]:
+                if hasattr(config, split) and getattr(config, split) is not None:
+                    _inject_collate(getattr(config, split))
+
+        _inject_collate(data_config)
 
     starter_config = ProjectConfig(
         model=model_component,
         data=data_config,
         trainer=trainer_component,
         output=output_config,
+    )
+
+    # Update targets to point to the local project structure
+    _update_targets(
+        starter_config,
+        f"trainite.models.{model_spec.name}",
+        f"models.{model_spec.name}",
+    )
+    _update_targets(
+        starter_config,
+        f"trainite.datasets.{dataset_spec.name}",
+        f"datasets.{dataset_spec.name}",
     )
 
     dump_config(starter_config, project_dir / "config.yaml")
--- a/trainite/config/model.py
+++ b/trainite/config/model.py
@@ -14,6 +14,3 @@ class TransformerModelConfig(ComponentConfig):
     feedforward_dim: int = Field(default=128, gt=0)
     dropout: float = Field(default=0.1, ge=0.0, lt=1.0)
     max_seq_len: int = Field(default=128, gt=0)
-    inference_every_epochs: int | None = Field(default=None, gt=0)
-    inference_num_samples: int = Field(default=5, gt=0)
-    max_inference_steps: int = Field(default=16, gt=0)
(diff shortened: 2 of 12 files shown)


aaishwarymishra:
@vfdev-5 Hi, this is kinda my approach.

what the trainer excepts for inference:
1. The model should have a generate method.
2. Dataset should have decode method that decodes the string.
3. Dataset has extract_prompt which extracts labels and prompts from the dataset values.

aaishwarymishra:
okay, Hi

So I kinda went with this approach.
1. Tokenizer is optional during training as dataset may provide it if needs the outputs to be tokenized externally, or give pre tokenized output.
2. For inference its mandatory as we need raw text in that case.
3. Collate fn has been moved to model as its model specific task.
3. String reverse now only gives the raw strings simplyfying it.

trainite/config/model.py
```diff
@@ -14,3 +14,6 @@ class TransformerModelConfig(ComponentConfig):
     feedforward_dim: int = Field(default=128, gt=0)
     dropout: float = Field(default=0.1, ge=0.0, lt=1.0)
     max_seq_len: int = Field(default=128, gt=0)
+    inference_every_epochs: int | None = Field(default=None, gt=0)
+    inference_num_samples: int = Field(default=5, gt=0)
+    max_inference_steps: int = Field(default=16, gt=0)
```
TahaZahid05: Why is this in TransformoerModelConfig? shouldn't this be in pretrainer?

trainite/models/transformer.py
```diff
@@ -152,3 +169,134 @@ def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
...
+                next_token = torch.where(
+                    eos_mask, torch.tensor(eos_token_id, device=device), next_token
+                )
+            generated = torch.cat([generated, next_token], dim=-1)
+
+            if (
+                eos_token_id is not None
+                and next_token.squeeze(-1).eq(eos_token_id).all()
+            ):
+                break
```
TahaZahid05, line 242: This does not seem right to me and also confusing, can you explain please?

trainite/models/transformer.py
```diff
@@ -152,3 +169,134 @@ def forward(self, input_ids: torch.Tensor) -> torch.Tensor:
...
+
+            input_ids = full_seq[:-1]
+            labels = full_seq[1:].clone()
+
+            # Mask the prompt portion in target labels
+            labels[: len(src) - 1] = self.ignore_index
+
+            input_ids_list.append(
+                input_ids.flip(0)
+            )  # Reverse input for better memory locality
```
TahaZahid05: aren't we reversing for left padding? the comment is misleading.

trainite/trainers/pretrainer.py
```diff
@@ -38,16 +50,29 @@ def __init__(self, config: ProjectConfig) -> None:
...
+        self.inference_tokenizer: Any = None
 
         self.device: str | torch.device = self._resolve_device()
         self.epochs: int = config.trainer.epochs
         self.log_every_steps: int = config.trainer.log_every_steps
         self.grad_clip_norm: float | None = getattr(
             config.trainer, "grad_clip_norm", None
         )
+        self.inference_every_epochs: int | None = getattr(
+            config.trainer, "inference_every_epochs", None
```
TahaZahid05, line 62: yeah, so here you are referencing from trainer, but you defined the fields in the model.

trainite/config/model.py
aaishwarymishra: My bad, before it was in trainer config , I moved it to the pretrainer but by mistake put it in model.

trainite/models/transformer.py
aaishwarymishra: hmm I see one issue that needs to be fixed. But essentially what we are doing is basically taking the token with highest probability. Checking in generated which tokens already have eos as the last item. If they do we dont actually care about tokens after that so we put eos token for there predictions. And put the predicted values where last token was eos.

trainite/models/transformer.py
aaishwarymishra: Auto suggest lol

trainite/trainers/pretrainer.py
aaishwarymishra: My bad

aaishwarymishra:
@TahaZahid05 Hi the eos token was causing confusing when we were checking if the sequence was ended so i replaced it with a special token

trainite/config/trainer.py
```diff
@@ -6,4 +6,7 @@
 class PreTrainerConfig(TrainerConfig):
     epochs: int = Field(default=3, gt=0)
     log_every_steps: int = Field(default=10, gt=0)
+    inference_every_epochs: int | None = Field(default=None, gt=0)
+    inference_num_samples: int = Field(default=5, gt=0)
+    max_inference_steps: int = Field(default=16, gt=0)
```
TahaZahid05: I think `max_new_tokens` is a better suited name?

trainite/cli/init.py
```diff
@@ -373,25 +373,38 @@ def init_project(args: argparse.Namespace) -> None:
...
+    # Inject model collator into data config dataloaders
+    if model_spec.collate_fn_target:
+        collate_target = model_spec.collate_fn_target
+
+        def _inject_collate(config):
+            if hasattr(config, "dataloader") and config.dataloader is not None:
+                config.dataloader.collate_fn = ComponentConfig(_target_=collate_target)
+            for split in ["train", "val", "test"]:
+                if hasattr(config, split) and getattr(config, split) is not None:
+                    _inject_collate(getattr(config, split))
```
TahaZahid05:
```suggestion
            dataloader = getattr(config, "dataloader", None)
            if dataloader is not None:
                dataloader.collate_fn = ComponentConfig(_target_=collate_target)
            for split in ["train", "val", "test"]:
                split_config = getattr(config, split, None)
                if split_config is not None:
                    _inject_collate(split_config)
```

trainite/datasets/string_reverse.py
TahaZahid05, line 28:
```suggestion
"""A simple character-level tokenizer with a hardcoded universal vocabulary.

   Maps each character in UNIVERSAL_VOCAB to a unique integer ID.
   ID 0 is reserved as the <PAD> token.
   ID 1 is reserved as the <BOS> token.
   ID 2 is reserved as the <SEP> token.
   ID 3 is reserved as the <EOS> token.
   ID 4 is reserved as the <UNK> token.
   """
```

TahaZahid05:
```suggestion
"""Generates unique random strings and their reversals.

   Each sample is returned as a dictionary containing:
       - 'source_text': the original random string
       - 'target_text': the reversed string
"""
```

TahaZahid05:
@aaishwarymishra i think some of the suggestions are not appearing correctly in review. you can just do them manually and commit i guess. also, i don't see any new additions in pretrainer tests file? Won't we need to add some tests for that as well?

trainite/config/trainer.py
aaishwarymishra: I think max, kinda infers to the upper range , and we should use inference to be consistent with other paramters.

TahaZahid05: how about `inference_max_new_tokens`?

aaishwarymishra: why you want the name changes? like whats the problem with current one

TahaZahid05: We should avoid the word `steps` as that is used for batch iterations mostly. `tokens` is much more suited since we are literally limiting the length of the generated output.

aaishwarymishra:
> @aaishwarymishra i think some of the suggestions are not appearing correctly in review. you can just do them manually and commit i guess. also, i don't see any new additions in pretrainer tests file? Won't we need to add some tests for that as well?

test for what?

PR #43, opened 2026-06-09
improve tests

aaishwarymishra:


aaishwarymishra:
I think we can leave templates and readme for now as it is, we will make them new when we make it public.

trainite/trainers/pretrainer.py
```diff
@@ -492,10 +483,47 @@ def _validate_dataloader_for_inference(self, loader: DataLoader, name: str) -> N
...
+        )
+        if self.inference_every_epochs is None:
             return
 
+        self.inference_num_samples: int = getattr(
+            config.trainer, "inference_num_samples"
+        )
+        self.max_inference_new_tokens: int = getattr(
+            config.trainer, "max_inference_new_tokens"
+        )
```
TahaZahid05, line 498: On second thought, i think we have to pass a fallback value here or None. If user deletes the values from `config.yaml` then the default values defined in pydantic class won't work and it will give error.

trainite/trainers/pretrainer.py
```diff
@@ -492,10 +483,47 @@ def _validate_dataloader_for_inference(self, loader: DataLoader, name: str) -> N
...
+
+        if (
+            isinstance(self.inference_every_epochs, bool)
+            or isinstance(self.inference_num_samples, bool)
+            or isinstance(self.max_inference_new_tokens, bool)
+            or not isinstance(self.inference_every_epochs, int)
+            or not isinstance(self.inference_num_samples, int)
+            or not isinstance(self.max_inference_new_tokens, int)
+        ):
```
TahaZahid05: just use type(). it can catch bool

trainite/trainers/pretrainer.py
aaishwarymishra: I dont understand, we are never generating the trainer pydantic code. The inference parameter is not universal for very use case , Is autoregressive dependent.

TahaZahid05: Try generating a new experiment on your PR branch. Set `inference_every_epochs` to 1 and remove:
```
  inference_num_samples: 5
  max_inference_new_tokens: 16
```

you will see an error even though we had defined default values for these fields. I am pointing out this issue.

aaishwarymishra: That should be the case, users should define the values, also we generate these in yaml so they wont run into this problem without removing them.

TahaZahid05: If the values are omitted, shouldn't it be the case that the user does not want to use them?

aaishwarymishra: They can make inference_every_epoch, none or remove it for that.

TahaZahid05: hm, makes sense let's stick with this then.

TahaZahid05 (approved):

TahaZahid05:
@aaishwarymishra  LGTM. you can merge after solving conflict

PR #44, opened 2026-06-09
Integrate RoPE in transformer

TahaZahid05:
The PR aims to replace absolute positional encoding with RoPE along with tests, and template changes.

### DIFF
--- a/trainite/models/transformer.py
+++ b/trainite/models/transformer.py
@@ -24,27 +24,36 @@ class Tokenizer(Protocol):
     def decode(self, ids: list[int]) -> str: ...
 
 
-class PositionalEncoding(nn.Module):
-    def __init__(self, d_model: int, max_len: int, dropout: float = 0.1) -> None:
+class RotaryEmbedding(nn.Module):
+    def __init__(self, dim: int) -> None:
         super().__init__()
-        self.dropout = nn.Dropout(p=dropout)
+        self.dim = dim
+        inv_freq = 1.0 / (10000.0 ** (torch.arange(0, dim, 2).float() / dim))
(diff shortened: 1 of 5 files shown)


trainite/models/transformer.py
```diff
@@ -24,31 +24,67 @@ def encode(self, text: str) -> list[int]: ...
...
+        self.register_buffer(
+            "sin_cached", emb.sin().unsqueeze(0).unsqueeze(0), persistent=False
+        )
 
-        if d_model % 2 != 0:
-            raise ValueError("d_model must be even for sinusoidal positional encoding.")
+    def forward(
+        self, x: torch.Tensor, seq_len: int
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        if seq_len > self.max_seq_len:
```
aaishwarymishra, line 59: This code is kind of redundant as its also the one used in the init, I think we can have a common helper, or not support sequences above max_seq_len.

trainite/models/transformer.py
```diff
@@ -59,6 +95,7 @@ def __init__(self, embed_dim: int, num_heads: int, dropout: float = 0.1):
         self.out = nn.Linear(embed_dim, embed_dim, bias=False)
         self.dropout = nn.Dropout(p=dropout)
         self.dropout_p = dropout
+        self.rotary_emb = RotaryEmbedding(dim=self.head_dim, max_seq_len=max_seq_len)
```
aaishwarymishra: The rotary embedding layer is same right? it never changes or is learnable, so I think its more efficient to have an global encoding and we pass it to individual blocks instead of having multiple local copies.

aaishwarymishra (approved):

PR #47, opened 2026-06-10
Simplyfy inference setup

aaishwarymishra:


trainite/trainers/pretrainer.py
```diff
@@ -565,8 +536,6 @@ def _log_inference(self, engine: Engine, loader: DataLoader, name: str) -> None:
         eos_token_id = getattr(tokenizer, "eos_token_id", None)
 
         dataset = loader.dataset
-        if not isinstance(dataset, Sized):
```
TahaZahid05, line 545: Let's not remove this entirely as it would result in error. Let's add a warning instead to inform user that we are skipping inference as it is a sized dataset.

PR #52, opened 2026-06-13
refactor: enforce explicit attribute initialization in PreTrainer

TahaZahid05:
### Context
Previously, the `PreTrainer` class initialized several of its instance variables (such as training/validation/test loaders, inference configuration options, and metrics registry) via side-effecting helper methods (e.g. `set_loaders()`, `_setup_inference()`, and `_attach_metrics()`). This pattern made `__init__` harder to read and trace because attributes were created or mutated inside nested helper calls.

### Changes
To improve code clarity and ensure all initializations happen explicitly in the constructor, the following changes were made:
- **Refactored DataLoader Setup**: Renamed and restructured `set_loaders` into `_build_dataloaders()`, which takes the data configuration and returns a tuple of training, validation, and test loaders, without mutating `self` directly.
- **Refactored Inference Setup**:
  - Assigned `inference_every_epochs`, `inference_num_samples`, and `max_inference_new_tokens` directly from `config.trainer` in `__init__`, avoiding redundant fallbacks.
  - Updated `_setup_inference()` to perform validation and return the resolved tokenizer instead of setting instance variables as side effects.
  - Corrected the conditional check for inference setup to use `is not None` to handle `0` epochs training properly.
- **Refactored Metrics Setup**: Updated `_attach_metrics()` to return the dictionary of attached metrics to be explicitly assigned to `self.metrics` in `__init__`.
- **Constructor Cleanup**: Removed pre-declarations of attributes in `__init__` and replaced them with sequential, explicit variable assignments.

### Verification
- **Unit Tests**: Ran all 74 unit tests via `pytest` (all passed).
- **Integration Test**: 
  - Initialized a starter template project using `trainite init my-experiment --model transformer --dataset string-reverse --trainer pretrainer --force --yes`.
  - Ran the generated training script (`python my-experiment/main.py my-experiment/config.yaml`) with inference logging enabled (`inference_every_epochs: 1`), verifying training, validation evaluation, metric calculation, and inference samples output successfully without regressions.

### DIFF
--- a/trainite/trainers/pretrainer.py
+++ b/trainite/trainers/pretrainer.py
@@ -75,21 +75,23 @@ class PreTrainer:
         torch.manual_seed(config.seed)
         self.config: ProjectConfig = config
 
-        self.train_loader: DataLoader
-        self.val_loader: DataLoader | None = None
-        self.test_loader: DataLoader | None = None
-        self.inference_tokenizer: object | None = None
-
         self.device: str | torch.device = self._resolve_device()
         self.epochs: int = config.trainer.epochs
         self.grad_clip_norm: float | None = getattr(
             config.trainer, "grad_clip_norm", None
         )
-        self.set_loaders()
+        self.train_loader, self.val_loader, self.test_loader = self._build_dataloaders(
+            config.data
+        )
         self.vocab_size: int = self._resolve_vocab_size()
         self.model: nn.Module = self._build_model()
 
-        self._setup_inference(config)
+        self.inference_every_epochs = config.trainer.inference_every_epochs
+        self.inference_num_samples = config.trainer.inference_num_samples
+        self.max_inference_new_tokens = config.trainer.max_inference_new_tokens
+        self.inference_tokenizer = (
+            self._setup_inference() if self.inference_every_epochs is not None else None
+        )
 
         self.loss_fn: nn.CrossEntropyLoss = nn.CrossEntropyLoss()
         self.optimizer: torch.optim.Optimizer = self._build_optimizer()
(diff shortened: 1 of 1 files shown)


trainite/trainers/pretrainer.py
```diff
@@ -209,30 +210,34 @@ def _build_loaders_from_ratios(
...
 
-    def set_loaders(self) -> None:
-        if isinstance(self.config.data, DataConfigBase):
-            self.train_loader = self._build_dataloader(self.config.data.train)
-            self.val_loader = (
-                self._build_dataloader(self.config.data.val)
-                if self.config.data.val
-                else None
+    def _build_dataloaders(
+        self, data_config: DataConfigBase | DataWithAutoSplit
```
aaishwarymishra: I think we should just use self.config , instead of unnecessary parameter passing.

trainite/trainers/pretrainer.py
```diff
@@ -305,11 +310,12 @@ def _eval_step(
         logits = self.model(inputs)
         return {"logits": logits, "targets": targets}
 
-    def _attach_metrics(self) -> None:
+    def _attach_metrics(self) -> dict[str, Loss | Accuracy]:
```
aaishwarymishra: We can define return as standard ignite metric or any, so user can easily add his/metric without needing to touch the union.

trainite/trainers/pretrainer.py
```diff
@@ -209,30 +210,34 @@ def _build_loaders_from_ratios(
...
-            self.train_loader = self._build_dataloader(self.config.data.train)
-            self.val_loader = (
-                self._build_dataloader(self.config.data.val)
-                if self.config.data.val
-                else None
+    def _build_dataloaders(
+        self, data_config: DataConfigBase | DataWithAutoSplit
+    ) -> tuple[DataLoader, DataLoader | None, DataLoader | None]:
+        train_loader: DataLoader
+        val_loader: DataLoader | None = None
```
aaishwarymishra: Hmm might need to check in with victor for this.

aaishwarymishra (approved):

PR #53, opened 2026-06-13
refactored test anand data config

aaishwarymishra:
## Description

  This PR refactors the automatic dataset splitting configuration parameters and fixes failures in the test
  suite introduced by the recent configuration schema changes.

  ### Key Changes

  1. Auto-Split Configuration Redesign:
      • Renamed  train_ratio  and  val_ratio  to  test_ratio  (default:  0.2 ) and  val_ratio  (default:  0.1 ).
      • Semantics changed so that both parameters are fractions of the total dataset, with the training set
      taking the remainder.
      • Enforced Pydantic validations ensuring the derived training ratio is between  0.5  and  0.9 .
  2. Trainer & Dataset Integration:
      • Updated data splitting logic in pretrainer.py.
      • Updated default configuration in dataset string_reverse.py.
      • Standardized config schemas in example code ().
  3. Test Suite Updates:
      • Updated base_test.py to validate the new  DataWithAutoSplit  schema constraints.
      • Refactored pretrainer_test.py to test auto-splits with the new ratio parameters.
      • Fixed tests verifying trainer execution without a validation loader by using  object.__setattr__  to
      bypass validation assignment errors on models where  val  is schema-required.				
  4. Documentation:
      • Updated root and example  README.md  files to align with the new ratio configuration keys.

### DIFF
--- a/tests/trainers/pretrainer_test.py
+++ b/tests/trainers/pretrainer_test.py
@@ -312,9 +312,7 @@ def test_pretrainer_run_with_val(project_config, temp_run_dir):
     assert "checkpoint_best" in trainer.handlers
 
 
-@pytest.mark.skip(
-    reason="Skipping this test because validation is required for early stopping and best checkpoint. "
-)
+@pytest.mark.skip(reason="Skipping this test because validation is required.")
 def test_pretrainer_run_without_val(project_config, temp_run_dir):
     object.__setattr__(
         project_config,
@@ -381,6 +379,7 @@ def test_pretrainer_test_method(project_config, temp_run_dir):
     assert "token_accuracy" in trainer.test_evaluator.state.metrics
 
 
+@pytest.mark.skip(reason="Skipping this test because validation is required ")
 def test_pretrainer_test_without_val(project_config, temp_run_dir):
     # Remove validation split
     # Add test split
--- a/tests/trainers/pretrainer_test.py
+++ b/tests/trainers/pretrainer_test.py
@@ -12,6 +12,7 @@ from trainite.config import (
     ComponentConfig,
     DataConfigBase,
     DataLoaderConfig,
+    DataWithAutoSplit,
     OptimizerConfig,
     OutputConfig,
     SplitConfig,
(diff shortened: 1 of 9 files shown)


tests/trainers/pretrainer_test.py
```diff
@@ -312,7 +313,15 @@ def test_pretrainer_run_with_val(project_config, temp_run_dir):
 
 
 def test_pretrainer_run_without_val(project_config, temp_run_dir):
```
TahaZahid05, line 316: This is confusing. In pydantic validation, we have restricted the user to have a val split (else there will be error), but we also have test checking if pretrainer works without val?

Why are we restrcting there to be a val split all the times? There can be scenarios where user does not want a val split?

trainite/trainers/pretrainer.py
```diff
@@ -218,7 +210,7 @@ def _build_loaders_from_ratios(
         return train_loader, val_loader, test_loader
 
     def set_loaders(self) -> None:
-        if self.config.data.train:
+        if isinstance(self.config.data, DataConfigBase):
```
TahaZahid05, line 213: you may have to update this pr since my pr also modifies some of these functions

trainite/trainers/pretrainer.py
```diff
@@ -230,22 +222,17 @@ def set_loaders(self) -> None:
...
+                "Automatic splitting requested with ratios: train=%s, val=%s, test=%s",
+                1.0 - self.config.data.test_ratio - self.config.data.val_ratio,
                 self.config.data.val_ratio,
+                self.config.data.test_ratio,
             )
             self.train_loader, self.val_loader, self.test_loader = (
                 self._build_loaders_from_ratios(self.config.data)
             )
 
-        if self.val_loader is None:
```
TahaZahid05, line 243: again, why are we making validation necessary?

TahaZahid05:
@aaishwarymishra my main worry is regarding making val necessary. I think only train split shall be necessary.

Also, kindly remove the changes in examples please. Let us not modify examples in every PR since it makes the PR and commit messy. We can regenerate examples in a seperate PR whenever we want to.

tests/trainers/pretrainer_test.py
aaishwarymishra: hmm, I will remove it

aaishwarymishra:
@TahaZahid05 As for the meeting , victor said val should be required for now.
If users want to make val optional we can do it in future.

Also about those val checks, I think those should have a separate pr for themselves as much of the code can be simplified.

tests/trainers/pretrainer_test.py
```diff
@@ -371,16 +383,23 @@ def test_pretrainer_test_method(project_config, temp_run_dir):
 
 def test_pretrainer_test_without_val(project_config, temp_run_dir):
```
TahaZahid05, line 383: add skip here as well

PR #54, opened 2026-06-14
Huggingface style

aaishwarymishra:
Summary

 Makes the Pretrainer/TransformerModel interface cleaner by establishing clear boundaries: padding lives in the
 trainer, generation lives in the model. Also makes the tokenizer mandatory on the model class.

 Changes

 ### trainite/trainers/pretrainer.py

 _resolve_tokenizer — simplified. Only checks for a class-level tokenizer attribute on the model. Raises immediately if
 missing. The old fallback getattr(self.model, "tokenizer", None) was always unreachable (model isn't built yet when
 this runs) and only worked because the bare except swallowed the AttributeError.

 _log_inference — padding is now done here with efficient tensor operations:

 ```python
   # reverse → pad_sequence → reverse (all tensor ops, no Python loops)
   input_ids = pad_sequence(reversed_tensors, ...).flip(1)
   attention_mask = input_ids.ne(pad_id).long()
 ```

 Passes a proper tensor + attention_mask to model.generate(), matching the HuggingFace convention. Also restored
 HF-compatible unwrapping (output.sequences if hasattr else output) so both raw-tensor and .sequences-style returns
 work.

 Removed dead if tokenizer is None guards throughout — tokenizer is now guaranteed present.

 ### trainite/models/transformer.py

 generate() — clean, minimalist signature:

 ```python
   def generate(
       self,
       input_ids: torch.Tensor,      # pre-padded, left-aligned
       max_new_tokens: int,
       attention_mask: torch.Tensor | None = None,
       eos_token_id: int | None = None,
       **kwargs,
   ) -> torch.Tensor:
 ```

 No list support, no pad_token_id — the caller handles padding. This is the same shape as HF's generate(). 

 Removed unused Protocol import.

 ### Tests

 - project_config fixture and from_ratios tests now use SimpleModelWithTokenizer (has class-level tokenizer)
 - GenerativeModel gains class-level tokenizer = DummyTokenizer()
 - test_setup_inference_missing_tokenizer updated to match new error message


trainite/models/transformer.py
```diff
@@ -172,18 +331,24 @@ def forward(
...
         num_heads: int = 2,
         feedforward_dim: int = 128,
         dropout: float = 0.1,
         max_seq_len: int = 128,
-        **kwargs,
+        **kwargs: Any,
     ) -> None:
         super().__init__()
+        self.tokenizer = CharTokenizer()
+        if vocab_size is None or vocab_size == 32:
```
TahaZahid05: why `vocab_size == 32`?

trainite/trainers/pretrainer.py
```diff
@@ -110,14 +100,37 @@ def _resolve_device(self) -> str | torch.device:
...
         while hasattr(dataset, "dataset"):
             dataset = getattr(dataset, "dataset")
         return dataset
 
     def _resolve_vocab_size(self) -> int:
-        train_dataset = self._get_underlying_dataset(self.train_loader.dataset)
-        vocab_size = getattr(train_dataset, "vocab_size", 0)
+        vocab_size = 0
+        if self.tokenizer is not None and hasattr(self.tokenizer, "vocab_size"):
+            vocab_size = getattr(self.tokenizer, "vocab_size", 0)
```
TahaZahid05: why need getattr?

trainite/trainers/pretrainer.py
```diff
@@ -110,14 +100,37 @@ def _resolve_device(self) -> str | torch.device:
...
+            train_dataset = self._get_underlying_dataset(self.train_loader.dataset)
+            vocab_size = getattr(train_dataset, "vocab_size", 0)
+
+        if vocab_size <= 0:
+            try:
+                target_symbol = get_target(self.config.model.target)
+                if hasattr(target_symbol, "vocab_size"):
+                    vocab_size = getattr(target_symbol, "vocab_size")
+            except Exception:
+                pass
```
TahaZahid05: can it be merged into one?

trainite/trainers/pretrainer.py
```diff
@@ -110,14 +101,37 @@ def _resolve_device(self) -> str | torch.device:
             resolved = "cuda" if torch.cuda.is_available() else "cpu"
         return resolved
 
+    def _resolve_tokenizer(self) -> Any:
+        model_class = get_target(self.config.model.target)
+        tokenizer = getattr(model_class, "tokenizer", None)
+        if tokenizer is None:
+            raise ValueError(
+                "Failed to resolve tokenizer from model. Please ensure your model class has a 'tokenizer' attribute or specify a tokenizer in your config."
```
```suggestion
                "Failed to resolve tokenizer from model. Please ensure your model class has a 'tokenizer' attribute."
```

trainite/trainers/pretrainer.py
```diff
@@ -110,14 +101,37 @@ def _resolve_device(self) -> str | torch.device:
...
+            )
+        return tokenizer
```
TahaZahid05: This feel a bit hacky. Creating tokenizer using the model before the model is even instantiated. Why not make tokenizer  part of `ProjectConfig`?

trainite/models/transformer.py
```diff
@@ -172,18 +331,24 @@ def forward(
 
 
 class TransformerModel(nn.Module):
+    tokenizer = CharTokenizer()
```
TahaZahid05: Again, this seems not a good design. We are setting a default tokenzier for the model while also accepting `input_ids` meaning that model should not need to know about tokenizer?

trainite/models/transformer.py
```diff
@@ -172,18 +331,24 @@ def forward(
...
         max_seq_len: int = 128,
-        **kwargs,
+        **kwargs: Any,
     ) -> None:
         super().__init__()
+        self.tokenizer = CharTokenizer()
+        if vocab_size is None or vocab_size == 32:
+            vocab_size = self.tokenizer.vocab_size
+
         self.embedding = nn.Embedding(vocab_size, hidden_size, padding_idx=0)
```
TahaZahid05: we can get padding index from tokenizer rather than hardcoding to 0.

trainite/trainers/pretrainer.py
```diff
@@ -109,35 +101,45 @@ def _resolve_device(self) -> str | torch.device:
...
+            if "vocab_size" in sig.parameters:
+                has_vocab_size_param = True
+        except Exception:
+            pass
+
+        kwargs = {}
+        if has_vocab_size_param:
+            kwargs["vocab_size"] = self.vocab_size
+
+        kwargs["pad_token_id"] = self.tokenizer.pad_token_id
```
TahaZahid05: perform the same signature check on pad_token_id as well as you did for vocab_size to prevent error for custom models.

trainite/trainers/pretrainer.py
```diff
@@ -544,13 +544,13 @@ def _setup_inference(self) -> object:
                 "Model must implement 'generate' method for inference logging."
             )
 
-        # Retrieve and check training tokenizer
-        train_ds = self._get_underlying_dataset(self.train_loader.dataset)
-        if not hasattr(train_ds, "tokenizer"):
+        tokenizer: Any = self.tokenizer
+        if not hasattr(tokenizer, "apply_chat_template") and not hasattr(
+            tokenizer, "decode"
+        ):
```
TahaZahid05: Conditional is incorrect. `_log_inference` requires `decode` strictly and requires either one of `apply_chat_template`, `encode` or the callable.

tests/models/transformer_test.py
```diff
@@ -153,63 +153,63 @@ def test_transformer_model_generate():
...
-
-        generated = model.generate(
-            ["ab"],
-            max_new_tokens=10,
-            tokenizer=tokenizer,
-        )
-        assert generated[0] == ""
+        assert generated[0].tolist() == [5, 6, tokenizer.eos_token_id]
+
+    # # Test with eos_token_id early exit (default tokenizer fallback)
```
TahaZahid05, line 189: remove this test

tests/trainers/pretrainer_test.py
```diff
@@ -269,10 +308,15 @@ def test_pretrainer_vocab_size_mismatch(project_config):
...
     project_config.model = cc(**model_conf)
 
-    with pytest.raises(ValueError, match="is smaller than the dataset vocabulary size"):
+    with pytest.raises(
+        ValueError, match="is smaller than the tokenizer vocabulary size"
+    ):
         PreTrainer(project_config)
 
 
+@pytest.mark.skip(
```
TahaZahid05: we can just remove.

tests/trainers/pretrainer_test.py
```diff
@@ -598,14 +643,28 @@ def test_setup_inference_missing_generate(project_config):
         PreTrainer(project_config)
 
 
+@pytest.mark.skip(reason="Obsolete after tokenizer became required in ProjectConfig")
```
TahaZahid05, line 651: same

TahaZahid05:
@aaishwarymishra I thought we wanted tokenizer in seperate folder?

trainite/trainers/pretrainer.py
aaishwarymishra: padding id should be mandatory

tests/trainers/pretrainer_test.py
aaishwarymishra: let it stay we can remove them once the structure is finalized

PR #58, opened 2026-06-17
refactor for tokenizer

aaishwarymishra:
Summary

Fix tests and import wiring to match the new trainite/tokenizers/ directory and updated dataset API. The tokenizer was extracted into its own package and
datasets now receive a tokenizer parameter at construction time — existing tests needed updating to reflect this.

Changes

New files
trainite/tokenizers/init.py — Package init exporting CharTokenizer, CharTokenizerConfig, UNIVERSAL_VOCAB
tests/tokenizers/char_test.py — 12 dedicated tokenizer tests covering encoding, decoding, special tokens, unk handling, batching, padding, truncation,
tensor conversion, and config
Test fixes
tests/datasets/string_reverse_test.py

Removed test_char_tokenizer (moved to tokenizer test file)
Fixed imports: CharTokenizer from trainite.tokenizers.char (was trainite.models.transformer), CHARSET_PRESETS from trainite.datasets.string_reverse
All StringReverseDataset constructors now pass the required tokenizer= parameter
Updated getitem assertions to match new tensor-returning API (input_ids/labels/attention_mask instead of prompt/completion)
Fixed stale -100 ignore-index filtering (labels no longer contain -100 at the dataset level)
Added 5 new tests: test_getitem_tensor_structure, test_get_item_inference, test_charset_empty_raises, test_dataset_config_min_seq_len_gt_max,
test_data_config_defaults
tests/models/transformer_test.py

Fixed CharTokenizer import path
Updated test_causal_lm_collate_fn to use pre-tokenized input_ids/labels dicts matching the new StringReverseDataset.getitem output format
tests/trainers/pretrainer_test.py

SimpleModel.forward accepts attention_mask kwarg (trainer now passes it)
SimpleDataset, SimpleDatasetNoVocab, EmptyDataset accept **kwargs (datasets now always receive tokenizer=)
GenerativeDataset implements get_item_inference (required by inference logging)
Updated error match strings for validation messages
tests/cli_test.py

Expected files list uses {trainer_spec.name}.py and {tokenizer_spec.name}.py instead of hardcoded names
Added get_tokenizer_spec, get_trainer_spec imports


trainite/trainers/pretrainer.py
```diff
@@ -414,74 +392,63 @@ def _setup_inference(self) -> object:
...
 
-        eos_token_id = getattr(tokenizer, "eos_token_id", None)
+        eos_token_id = self.tokenizer.eos_token_id
+        sep_token_id = self.tokenizer.sep_token_id
+        bos_token_id = self.tokenizer.bos_token_id
+        pad_token_id = getattr(self.tokenizer, "pad_token_id", 0)
 
         dataset = loader.dataset
-        if not isinstance(dataset, Sized):  # type: ignore
-            raise TypeError(f"{name} dataset must be Sized to run inference logging.")
```
TahaZahid05, line 427: i think it is better to keep this as it is more informative.

trainite/trainers/pretrainer.py
```diff
@@ -414,74 +392,63 @@ def _setup_inference(self) -> object:
...
-            prompts.append(prompt)
-            targets.append(target)
 
+        # Extract prompt IDs via dataset.get_item_inference (validated in _setup_inference).
+        prompt_ids_list: list[torch.Tensor] = []
+        prompt_attn_list: list[torch.Tensor] = []
+        targets: list[str] = []
+        sources: list[str] = []
+        for i in range(num_samples):
+            dataset = dataset.dataset if isinstance(dataset, Subset) else dataset  # in case of Subset wrapping
```
TahaZahid05: We don't need to look at the underlying dataset? This actually breaks logic because you are picking indices from original dataset rather than the split?

trainite/models/transformer.py
```diff
@@ -349,79 +178,17 @@ def forward(self, input_ids: torch.Tensor, attention_mask: torch.Tensor | None =
         B, S = input_ids.shape
         x = self.embedding(input_ids) * math.sqrt(self.embedding.embedding_dim)
         cos, sin = self.rotary_emb(x, seq_len=S)
+        padding_mask: torch.Tensor | None = None
         if attention_mask is not None:
-            padding_mask = attention_mask.reshape(B, 1, 1, S).to(torch.bool)
-        elif (input_ids == self.embedding.padding_idx).any():
+            if not attention_mask.any().item():
```
TahaZahid05:
```suggestion
            if not attention_mask.all().item():
```

current check is seeing if none of tokens are being attended to rather than if atleast one is not being attended to so it can apply padding.

trainite/shared/main.py
```diff
@@ -151,7 +172,7 @@ def main() -> None:
     config = load_config(config_path, ProjectConfig)
 
     device = resolve_device(config.device)
-    tokenizer = instantiate(config.tokenizer)
+    tokenizer = instantiate(config.preprocessor)
```
TahaZahid05:
```suggestion
    tokenizer = instantiate(config.preprocessor) if config.preprocessor is not None else None
```

trainite/cli/init.py
```diff
@@ -181,8 +176,13 @@ def _build_templates(model_name: str, dataset_name: str, trainer_name: str, proj
     model_spec = get_model_spec(model_name)
     dataset_spec = get_dataset_spec(dataset_name)
     trainer_spec = get_trainer_spec(trainer_name)
+    preprocessor_spec = (
+        get_preprocessor_spec(dataset_spec.preprocessor_spec_name) if dataset_spec.preprocessor_spec_name else None
+    )
+    specs = [model_spec, dataset_spec, trainer_spec]
+    specs = specs + [preprocessor_spec] if preprocessor_spec else specs
```
TahaZahid05:
```suggestion
    specs = [model_spec, dataset_spec, trainer_spec]
    preprocessor_spec = None
    if dataset_spec.preprocessor_spec_name:
        preprocessor_spec = get_preprocessor_spec(dataset_spec.preprocessor_spec_name)
        specs.append(preprocessor_spec)
```

trainite/cli/init.py
```diff
@@ -249,14 +273,30 @@ def _build_templates(model_name: str, dataset_name: str, trainer_name: str, proj
...
 
+    if preprocessor_spec:
+        templates[f"preprocessors/{preprocessor_spec.name}.py"] = _render_template(
+            PROJECT_ROOT / preprocessor_spec.implementation_path,
+            preprocessor_spec.template_replacements,
+        )
+
+    templates.update(
+        {
+            f"{trainer_spec.name}.py": _render_template(
```
TahaZahid05: should we really keep generated file name `pretrainer.py`? I think `trainer.py` seems better suited for a fresh generated project? You also would have to dynamically change this file name in markdown/templates based on trainer selected if not.

trainite/datasets/string_reverse.py
```diff
@@ -22,8 +24,8 @@ class StringReverseDataset(Dataset):
     """Generates unique random strings and their reversals.
 
     Each sample is returned as a dictionary containing:
-        - 'prompt': the original random string
-        - 'completion': the reversed string
+        - 'source': the original random string
```
TahaZahid05, line 27: check if all markdowns are also updated.

trainite/datasets/string_reverse.py
```diff
@@ -89,27 +108,65 @@ def __init__(
             # Pack each unique sequence into source and target formats
```
TahaZahid05, line 89: `per_seq_size` seemed more descriptive

trainite/trainers/pretrainer.py
```diff
@@ -346,36 +334,29 @@ def _run_evaluations(self, engine: Engine) -> None:
...
                 train_metrics["token_accuracy"],
             )
 
     def _validate_dataloader_for_inference(self, loader: DataLoader, name: str) -> None:
         dataset = loader.dataset
-        if not isinstance(dataset, Sized) or len(dataset) == 0:  # type: ignore
-            return
-
-        item = dataset[0]
+        dataset = dataset.dataset if isinstance(dataset, Subset) else dataset
```
TahaZahid05, line 356: again, why checking overall dataset? what if the split is empty but overall dataset is not?

Also, i think the prior check was more readable as it explicitly checked `len(dataset)`

trainite/trainers/pretrainer.py
```diff
@@ -399,12 +380,9 @@ def _setup_inference(self) -> object:
...
             )
 
-        if not hasattr(self.model, "generate"):
-            raise ValueError("Model must implement 'generate' method for inference logging.")
-
         tokenizer: Any = self.tokenizer
-        if not hasattr(tokenizer, "apply_chat_template") and not hasattr(tokenizer, "decode"):
-            raise ValueError("Tokenizer must implement 'apply_chat_template' or 'decode' method for inference logging.")
+        if not hasattr(tokenizer, "decode"):
+            raise ValueError("Tokenizer must implement 'decode' method for inference logging.")
```
```suggestion
        if not hasattr(tokenizer, "decode") or not callable(tokenizer):
            raise ValueError("Tokenizer must be callable and implement 'decode' method for inference logging.")
```

trainite/trainers/pretrainer.py
aaishwarymishra: That was just to please the linter. Not actually necessary

trainite/cli/init.py
aaishwarymishra: I think it should be dynamically generated, also we have to rename the pretrainer to a new name.

trainite/trainers/pretrainer.py
```diff
@@ -351,6 +351,8 @@ def _run_evaluations(self, engine: Engine) -> None:
 
     def _validate_dataloader_for_inference(self, loader: DataLoader, name: str) -> None:
         dataset = loader.dataset
+        if len(dataset) == 0:
+            raise ValueError(f"{name} dataset is empty. Cannot perform inference logging.")
```
TahaZahid05, line 355: maybe, warning would be better as on small dataset splits can become 0. You also may have to skip in `_log_inference` if samples are 0

trainite/cli/init.py
TahaZahid05: hm, ok. what would you suggest as new name? Also, if it should be dynamically generated, then markdowns need to be updated as well.

trainite/trainers/pretrainer.py
aaishwarymishra: This check is on the setup or before training even begins so trowing error will be fine.

PR #61, opened 2026-06-21
Generalize decoder trainer: move inference prompt format into the transform

aaishwarymishra:
Summary

PreTrainer is meant to be a generic decoder trainer, but _log_inference hardcoded the prompt layout [bos] + source + [sep] — which is really the StringReverse dataset's format (the prompt half of PromptCompletionTransform.__call__). The "generic" trainer secretly knew one dataset's token scheme, so any dataset with a different prompt structure would silently get the wrong prompt.

This moves prompt construction to its rightful owner — the transform — and hands it to the trainer by dependency injection. The trainer now knows nothing dataset-specific; it only keeps generic decoder concerns (pad_token_id for batching, eos_token_id for stopping, decode for display).

Changes

- datasets/string_reverse.py — add PromptCompletionTransform.build_prompt(sample) -> list[int] (the prompt-only token ids); __call__ now reuses it, so the format has a single source of truth. Signature takes the full sample dict to stay consistent with __call__ and to support future multi-field prompts (e.g. chat templates).
- shared/main.py — pulls the transform off the train loader (unwrapping the random_split Subset) and injects it as prompt_transform= into PreTrainer.
- trainers/pretrainer.py — inference builds prompts via the injected transform instead of assembling special tokens itself; _setup_inference validates the transform exposes build_prompt; removed a stale comment referencing a non-existent dataset.get_item_inference.

### DIFF
--- a/trainite/trainers/pretrainer.py
+++ b/trainite/trainers/pretrainer.py
@@ -361,7 +361,7 @@ class PreTrainer:
         if not hasattr(tokenizer, "decode") or not callable(tokenizer):
             raise ValueError("Tokenizer must be callable and implement 'decode' method for inference logging.")
 
-        # Validate active loaders yield source/target
+        # Validate active loaders
         self._validate_dataloader_for_inference(self.train_loader, "Train")
         if self.val_loader is not None:
             self._validate_dataloader_for_inference(self.val_loader, "Val")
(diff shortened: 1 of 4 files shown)


trainite/trainers/pretrainer.py
```diff
@@ -354,60 +350,40 @@ def _validate_dataloader_for_inference(self, loader: DataLoader, name: str) -> N
...
-                f"max_inference_new_tokens={self.max_inference_new_tokens}"
-            )
-
+    def _setup_inference(self) -> None:
         tokenizer: Any = self.tokenizer
         if not hasattr(tokenizer, "decode") or not callable(tokenizer):
             raise ValueError("Tokenizer must be callable and implement 'decode' method for inference logging.")
 
-        # Validate active loaders
+        # Validate active loaders yield source/target
```
TahaZahid05: No need to update comment. Revert please.

TahaZahid05 (approved):

PR #65, opened 2026-06-23
Added attention matrix support for experiment

TahaZahid05:
The PR aims to add support of manual attention calculation for when we need attention matrix for visualization as well. Implementation reference is taken from: [torch.nn.functional.scaled_dot_product_attention](https://docs.pytorch.org/docs/2.12/generated/torch.nn.functional.scaled_dot_product_attention.html)

It also makes the existing spda function calling code clean by reducing the spda call in the conditional.

### DIFF
--- a/examples/string-reverse-fixed/models/transformer.py
+++ b/examples/string-reverse-fixed/models/transformer.py
@@ -97,7 +97,8 @@ class Attention(nn.Module):
         cos: torch.Tensor,
         sin: torch.Tensor,
         padding_mask: torch.Tensor | None = None,
-    ) -> tuple[torch.Tensor, torch.Tensor]:
+        output_attentions: bool = False,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor | None]:
         B, S, C = x.shape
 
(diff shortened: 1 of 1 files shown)


aaishwarymishra:
I think it would be better if we make it a helper method and 1:1 replica of the provided code :)

examples/string-reverse-fixed/models/transformer.py
```diff
@@ -111,29 +112,85 @@ def forward(
...
+            context, attn_weights = scaled_dot_product_attention_reference(
                 q,
                 k,
                 v,
                 attn_mask=mask,
-                is_causal=False,
-                dropout_p=self.dropout_p if self.training else 0.0,
+                dropout_p=self.dropout_p,
+                is_causal=is_causal,
+                training=self.training,
```
aaishwarymishra, line 131: I dont we need extra training, parameter.

TahaZahid05: ig it's fine since we want to keep it the same as the reference pytorch implementation. The reference function just changes dropout based on training param in function rather than in calling

Issue #67, opened 2026-06-24
CLI issue with uninstalled deps


vfdev-5:
I could successfully install trainite with minimal deps (see #66) using `uv sync --no-dev`:
and can get the help output:
```bash
(trainite) /playground/transformer-experiments# uv run trainite -h
usage: /playground/trainite/.venv/bin/trainite [-h] [OPTIONS]

Generate a starter training project.

Run ``trainite init`` without any arguments to enter interactive mode, where each option is prompted one at a time.

╭─ positional arguments ───────────────────────────────────────────────────────────────────────╮
│ [STR]              Directory to create the starter project in. (default: my-cool-experiment) │
╰──────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ options ────────────────────────────────────────────────────────────────────────────────────╮
│ -h, --help         show this help message and exit                                           │
│ --model {transformer}                                                                        │
│                    Starter model template to use. (default: transformer)                     │
│ --dataset {string-reverse}                                                                   │
│                    Starter dataset template to use. (default: string-reverse)                │
│ --trainer {decoder-trainer}                                                                  │
│                    Starter trainer template to use. (default: decoder-trainer)               │
│ --output-root STR  Output root for generated config. (default: outputs)                      │
│ --run-name STR     Run name for generated config. (default: '')                              │
│ --force, --no-force                                                                          │
│                    Overwrite existing starter files. (default: False)                        │
│ -y, --yes, --no-yes                                                                          │
│                    Skip prompts and use defaults for anything not provided. (default: False) │
╰──────────────────────────────────────────────────────────────────────────────────────────────╯
```

However if I call it with inexisting attribute there is an unhandled error:
```bash
(trainite) /playground/transformer-experiments# uv run trainite abc
Traceback (most recent call last):                                            
  File "/playground/trainite/trainite/shared/utils.py", line 23, in get_target                                                                               
    module = importlib.import_module(module_path)                                                                                                            
  File "/root/.local/share/uv/python/cpython-3.13.12-linux-aarch64-gnu/lib/python3.13/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
[...]

PR #69, opened 2026-06-24
fix/ tyro random input taking

aaishwarymishra:
fixes the issue #67


trainite/cli/main.py
```diff
@@ -9,18 +9,17 @@
 def main(argv: Sequence[str] | None = None) -> None:
     args_list = list(argv) if argv is not None else sys.argv[1:]
 
-    if not args_list:
+    if not args_list or args_list[0] != "init":
```
vfdev-5, line 12: I would check for a list of available commands later instead of checking for hardcoded "init"

aaishwarymishra: To be honest we wont even need these checks in the future, these subcommands check is present right now because for subcommans tyro requires union and we only have 1 command right now..

Issue #74, opened 2026-06-29
Feedback on generated code


vfdev-5:
- [x] 1) Instead of https://github.com/pytorch-ignite/trainite/blob/bf6f4dfaf91f133ab0219777707454905e92c25f/trainite/shared/main.py#L162-L168
let's use ignite's https://github.com/pytorch/ignite/blob/a9c793e2fe8976f0bd92ced798f74251a1d74554/ignite/utils.py#L239

- [x] 2) Simplify device specification using `idist.device()`: 
Define as following 
```diff
class ProjectConfig(BaseModel):
    ...
    - device: str = "auto"
    + device: str | None = None
```
and in the code, remove `resolve_device` and do : 
```python
device = idist.device() if config.device is None else config.device
``` 

- [x] 3) If we have trainer abstraction I do not understand why we now put into shared/main.py all language model specific things.
We can make a single file specific code instead of two. 
We can make main.py file generic by creating the trainer. Trainer class should be defined in the config like the model or the optimizer and in main.py 
```python
def main() -> None:
    args = parse_args()
    config_path = Path(args.config)
    config = load_config(config_path, ProjectConfig)

    # Later we can insert the code for DDP: with idist.Parallel(...)
    trainer = instantiate(config.trainer, config=config)
    trainer.run()

if __name__ == "__main__":
    main()    
```

Let's decoder instantiate everything in the constructor in few steps:
- dataflow
- model, optimizer, lr scheduler, 
- engine and handlers

- [x] 4) Let's remove `class ComponentConfig(BaseModel):` and each Config class should inherit from pydantic BaseModel. No subclassing.

- [x] 5) Apply "separate concerns" logic in for dataflow creation. Instead of 
```python
    tokenizer = instantiate(config.preprocessor) if config.preprocessor is not None else None

    train_loader, val_loader, test_loader = build_dataloaders(config)

    vocab_size = resolve_vocab_size(tokenizer, config.model)  # <-- I do not think model config should contain vocab_size, we take it from the tokenizer.
    model = build_model(config.model, tokenizer, vocab_size, device)

    optimizer = instantiate(config.optimizer, params=model.parameters())

    # The transform owns the prompt format; hand it to the trainer for inference logging.
    ds = train_loader.dataset
    ds = ds.dataset if isinstance(ds, Subset) else ds
    prompt_transform = getattr(ds, "transform", None)
```
let's do the following:
```python
# This is the trainer constructor:
[...]

PR #76, opened 2026-06-30
Refactor DecoderTrainer Lifecycle, Handler Management, and Test Verification

aaishwarymishra:
## Summary

  This PR refactors the initialization lifecycle and handler configuration of  DecoderTrainer  to eliminate
  potential bugs, simplify the instance properties, resolve static typing warnings, and establish more robust
  test coverage.
  
  part of #74

### DIFF
--- a/trainite/trainers/decoder_trainer.py
+++ b/trainite/trainers/decoder_trainer.py
@@ -247,7 +247,7 @@ class DecoderTrainer:
 
     def setup_early_stopping(self) -> None:
         patience = self.config.trainer.early_stopping_patience
-        if self.val_loader and patience is not None:
+        if patience is not None:
             early_stopping = EarlyStopping(
                 patience=patience,
                 score_function=lambda engine: engine.state.metrics["loss"],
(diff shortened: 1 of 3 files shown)


trainite/trainers/decoder_trainer.py
```diff
@@ -84,23 +84,45 @@ def __init__(
...
+        # Attach checkpointing
+        self.checkpointers = self.setup_checkpointing()
+
+        # Attach TensorBoard logger
+        self.tb_logger = self._setup_tensorboard()
+
+        # Run evaluations at the end of each epoch to log training and validation metrics
+        self.engine.add_event_handler(Events.EPOCH_COMPLETED, self._run_evaluations)
+
+        # Attach inference logger if inference logging is enabled
```
TahaZahid05:
```suggestion
        # Setup inference and attach inference logger if inference logging is enabled
```

trainite/trainers/decoder_trainer.py
```diff
@@ -512,7 +538,7 @@ def test(self, test_loader: DataLoader | None = None) -> None:
             return
 
         # Load best model if available
-        checkpoint_handler = self.handlers.get("checkpoint_best") or self.handlers.get("checkpoint_last")
+        checkpoint_handler = self.checkpointers.get("checkpoint_best") or self.checkpointers.get("checkpoint_last")
```
TahaZahid05: This isn't correct. `checkpoint_best` always exist even when not ran meaning this condition will always evaluate to checkpoint_best

trainite/trainers/decoder_trainer.py
```diff
@@ -205,46 +228,54 @@ def _attach_handlers(self) -> None:
...
+        self.engine.add_event_handler(
+            Events.EPOCH_COMPLETED(every=self.inference_every_epochs),
+            self._log_inference,
+            self.val_loader,
+            "Val",
+        )
+
+    def setup_early_stopping(self) -> None:
+        patience = self.config.trainer.early_stopping_patience
+        if self.val_loader and patience is not None:
```
TahaZahid05: no need to check `self.val_loader` as we have made val split required.

trainite/trainers/decoder_trainer.py
TahaZahid05: here as well. let's remove this check since val split is now requried?

trainite/trainers/decoder_trainer.py
```diff
@@ -512,13 +529,16 @@ def test(self, test_loader: DataLoader | None = None) -> None:
...
+        checkpoint_handler = self.checkpointers.get("checkpoint_best", None)
         if checkpoint_handler and checkpoint_handler.last_checkpoint:
             checkpoint_path = checkpoint_handler.last_checkpoint
 
             self.logger.info("Loading best model for testing from %s", checkpoint_path)
             checkpoint = torch.load(checkpoint_path, map_location=self.device, weights_only=True)
             self.model.load_state_dict(checkpoint["model"])
+        else:
+            self.logger.warning("No best model checkpoint found. Using current model for testing.")
+            return
```
TahaZahid05:
```suggestion
```

aaishwarymishra: we would have to skip testing dont we if best model not found?

TahaZahid05: Not really, can be the case where model is loaded manually or if user is testing a fresh model.

TahaZahid05 (approved):

PR #78, opened 2026-07-01
Move component initialization into DecoderTrainer

aaishwarymishra:
Changes

- decoder_trainer.py — absorbs build_model, build_dataset, build_dataloaders, create_dataloader, resolve_vocab_size, _inject_if_accepted from main.py. __init__ is now config-only: DecoderTrainer(config). Dataloader construction is split into two clearly-typed helpers (_loaders_from_splits / _loaders_from_ratios) dispatched via isinstance instead of hasattr.
- shared/main.py — 186 lines → 15 lines. Load config, run trainer.
- registry.py — adds datasets.transformed to the decoder-trainer template replacements so generated trainer.py gets the correct local import.
- cli/init.py — removes the now-redundant TransformedDataset rewrite from main_replacements.
- Tests — create_trainer_from_config collapses to return DecoderTrainer(config).

Model/dataset contract

build_model uses _inject_if_accepted for both vocab_size and pad_token_id — models declare what they need in their __init__ signature and the trainer injects accordingly. Same pattern already used for datasets and transforms.


TahaZahid05:
@aaishwarymishra Kindly solve the merge conflict.

trainite/trainers/decoder_trainer.py
```diff
@@ -58,36 +61,143 @@ class ProjectConfig(BaseModel):
...
-    ) -> None:
+    def __init__(self, config: ProjectConfig) -> None:
         self.logger: logging.Logger = setup_logger("trainer", level=logging.INFO)
         torch.manual_seed(config.seed)
         self.config: ProjectConfig = config
-        self.tokenizer = preprocessor
-        self.prompt_transform = prompt_transform
-        self.device: str | torch.device = self._resolve_device()
+        self.device: str | torch.device = idist.device() if config.device == "auto" else config.device
+        self.tokenizer = instantiate(config.preprocessor)
```
TahaZahid05, line 185: tokenizer is optional so can be None as well.

trainite/trainers/decoder_trainer.py
```diff
@@ -58,36 +61,143 @@ class ProjectConfig(BaseModel):
...
+        return {}
+
+
+# Builds the model based on the provided configuration, tokenizer, and device.
+def build_model(model_config: Any, tokenizer: Any, vocab_size: int, device: str | torch.device) -> nn.Module:
+    target_symbol = get_target(model_config.target)
+    kwargs = _inject_if_accepted(
+        target_symbol,
+        vocab_size=vocab_size,
+        pad_token_id=tokenizer.pad_token_id,
```
TahaZahid05, line 99: same here

trainite/cli/init.py
```diff
@@ -29,7 +29,7 @@ class ProjectConfig(BaseModel):
     trainer: BaseModel
     output: OutputConfig
     seed: int = 42
-    device: str = "auto"
+    device: str = "cpu"
```
TahaZahid05: The device simplification code seems not be aligned with what Victor pointed out in feedback. Check the issue for reference. You may also see the implementation inside branch `simplify_device` as reference

trainite/trainers/decoder_trainer.py
aaishwarymishra: For decoder model or any language model tokenizer cant be None

TahaZahid05: Let's then update the typehint for preprocessor to only be `BaseModel`

trainite/cli/init.py
aaishwarymishra: Hi the thing is we should let the user decide which device he/she wants to use, so for auto we can detect already which device the person has, if he/she wants to use cpu they can specify it, during generation phase , we should use auto I will fix it.

TahaZahid05: `auto` doesn't seem to be pytorch standard as far as I know. If the user specifies device in config (cuda or cpu), we can use it else we can use `idist` to assign ourselves. Not sure that we should add `auto`

aaishwarymishra: let me ask victor

aaishwarymishra:
done

TahaZahid05 (approved):

aaishwarymishra:
part of #74

PR #80, opened 2026-07-01
Mock `torch` and `ignite` dynamically during CLI scaffolding

TahaZahid05:
## Description
This pull request resolves the `ModuleNotFoundError` that occurs when running `trainite init` in an environment without `torch` and `pytorch-ignite` installed.

Part of #74


trainite/__init__.py
aaishwarymishra, line 1: to me more clear we can make it as a separate file like like mocker.py and import it in trainite/cli/main.py and simply pass the list for the torch and ignite instead of checking explicitly with try block.

TahaZahid05:
@aaishwarymishra done.

trainite/cli/mocker.py
```diff
@@ -0,0 +1,48 @@
...
+        # parent module ('torch') so that dot-notation syntax does not crash.
+        if "." in module.__name__:
+            parent_name, child_name = module.__name__.rsplit(".", 1)
+            parent_module = sys.modules.get(parent_name)
+            if parent_module:
+                setattr(parent_module, child_name, module)
+
+
+def mock_dependencies(*packages):
+    import importlib
```
aaishwarymishra: it should be on top

TahaZahid05:
@aaishwarymishra done

aaishwarymishra:
@TahaZahid05 I meant the import statement not whole function

aaishwarymishra (approved):

PR #85, opened 2026-07-02
refactor: move code to utils.py

aaishwarymishra:
This pull request refactors and centralizes shared utility functions for dataset, dataloader, model setup, and Ignite engine wiring into a new `trainite/shared/utils.py` module. It updates the `Trainer` class in `decoder_trainer.py` and related tests to use these shared utilities, resulting in a cleaner and more maintainable codebase. The changes also improve code reuse and consistency across the project.

**Refactoring and Centralization of Utilities:**

* Moved all builder and Ignite engine setup functions (such as `build_model`, `build_dataset`, `create_dataloader`, `build_dataloaders`, `make_run_dir`, `attach_lr_scheduler`, `attach_early_stopping`, `setup_checkpointing`, `setup_experiment_logger`, `upload_model_to_wandb`, and `setup_console_logger`) from `decoder_trainer.py` into a new shared file, `trainite/shared/utils.py`. All related imports and usages were updated accordingly. [[1]](diffhunk://#diff-95664947e337a24cae1f3e4435b2f7c9583235285155de1ca6447fa7b50257bdR1-R32) [[2]](diffhunk://#diff-fac8db1e50d0671d7f20e3686be7e2b854d30ae6906a3f8b919154c5b760b20dL1-R35) [[3]](diffhunk://#diff-fac8db1e50d0671d7f20e3686be7e2b854d30ae6906a3f8b919154c5b760b20dL85-L181) [[4]](diffhunk://#diff-fac8db1e50d0671d7f20e3686be7e2b854d30ae6906a3f8b919154c5b760b20dL202-R141) [[5]](diffhunk://#diff-fac8db1e50d0671d7f20e3686be7e2b854d30ae6906a3f8b919154c5b760b20dL235-L241)

**Trainer Initialization and Logging Improvements:**

* The `Trainer` class in `decoder_trainer.py` now uses the shared utility functions for run directory creation, logger setup, learning rate scheduling, early stopping, checkpointing, and experiment logging, removing duplicated code and improving clarity. [[1]](diffhunk://#diff-fac8db1e50d0671d7f20e3686be7e2b854d30ae6906a3f8b919154c5b760b20dL202-R141) [[2]](diffhunk://#diff-fac8db1e50d0671d7f20e3686be7e2b854d30ae6906a3f8b919154c5b760b20dL235-L241)

**Test Refactoring:**

* Updated tests to import and use the new `build_dataset` and `upload_model_to_wandb` utilities from `trainite.shared.utils` instead of the old locations. This includes changes in both `tests/datasets/string_reverse_test.py` and `tests/trainers/decoder_trainer_test.py`. [[1]](diffhunk://#diff-c2f4f8b4321a41e2c5a0e98a2f02dcb6035a6ab905de64800706b537c137bcf4L12-R12) [[2]](diffhunk://#diff-3546af870c8a70aff1cd32740059f730c9a41560736d7e082573923b30016627R21) [[3]](diffhunk://#diff-3546af870c8a70aff1cd32740059f730c9a41560736d7e082573923b30016627L390-R396) [[4]](diffhunk://#diff-3546af870c8a70aff1cd32740059f730c9a41560736d7e082573923b30016627L402-R413)

**CLI Template Generation:**

* Updated the CLI initialization logic in `trainite/cli/init.py` to use a replacement mapping when rendering the shared `utils.py` template, ensuring correct import paths in generated projects.

These changes improve maintainability, encourage code reuse, and make it easier to add new trainers or datasets while keeping the codebase DRY and organized.

### DIFF
--- a/trainite/shared/utils.py
+++ b/trainite/shared/utils.py
@@ -363,7 +363,7 @@ def upload_model_to_wandb(
     run_name: str,
     logger: logging.Logger,
 ) -> None:
-    checkpoint = checkpointers.get("checkpoint_best") or checkpointers.get("checkpoint_last")
+    checkpoint = checkpointers.get("checkpoint_best")
     checkpoint_path = checkpoint.last_checkpoint if checkpoint else None
     if not checkpoint_path:
         logger.warning("No checkpoint found; skipping model upload to Weights & Biases.")
--- a/trainite/shared/utils.py
+++ b/trainite/shared/utils.py
@@ -1,10 +1,35 @@
+from datetime import datetime
 import importlib
+import inspect
+import logging
 from pathlib import Path
-from typing import Any, TypeVar
+from typing import Any, Literal, TypeVar
 
-import yaml
+from ignite.engine import Engine, Events
+from ignite.handlers import (
+    Checkpoint,
+    DiskSaver,
+    EarlyStopping,
+    create_lr_scheduler_with_warmup,
+)
+from ignite.handlers.fbresearch_logger import FBResearchLogger
+from ignite.handlers.param_scheduler import ParamScheduler
+from ignite.handlers.tensorboard_logger import TensorboardLogger
+from ignite.handlers.wandb_logger import WandBLogger
+from ignite.utils import setup_logger
 from omegaconf import OmegaConf
 from pydantic import BaseModel
+import torch
+from torch import nn
+from torch.optim.lr_scheduler import LinearLR
+from torch.utils.data import DataLoader, Dataset, random_split
+import yaml
+
+from trainite.config.base import (
+    DataConfigBase,
+    DataWithAutoSplit,
+)
+from trainite.datasets.transformed import TransformedDataset
 
 T = TypeVar("T", bound=BaseModel)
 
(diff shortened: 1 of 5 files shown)


trainite/trainers/decoder_trainer.py
```diff
@@ -199,31 +99,46 @@ def __init__(self, config: ProjectConfig) -> None:
...
-        self.logger = self.attach_loggers()
+        self.logger = setup_console_logger(self.run_dir, self.engine, config.trainer.log_every_steps, self.optimizer)
 
         # Attach learning rate scheduler
-        self.attach_lr_scheduler()
+        self.scheduler = attach_lr_scheduler(self.engine, self.optimizer, self.total_iters, config.optimizer.lr)
 
         # Attach early stopping
-        self.setup_early_stopping()
+        attach_early_stopping(self.val_evaluator, self.engine, config.trainer.early_stopping_patience)
```
TahaZahid05, line 119: why are we not saving early stopping? `attach_early_stopping` is returning the object

trainite/shared/utils.py
```diff
@@ -72,3 +97,301 @@ def dump_config(config: BaseModel, path: str | Path) -> None:
...
+    return exp_logger
+
+
+def upload_model_to_wandb(
+    exp_logger: Any,
+    checkpointers: dict[str, Checkpoint],
+    run_name: str,
+    logger: logging.Logger,
+) -> None:
+    checkpoint = checkpointers.get("checkpoint_best") or checkpointers.get("checkpoint_last")
```
TahaZahid05: I think we discussed this in a previous PR as well that the first condition will always be true which makes this not correct. Can you check?

trainite/trainers/decoder_trainer.py
aaishwarymishra: Because we never need to interact with it

trainite/shared/utils.py
aaishwarymishra: Hmm true but I think its more clear and safe to do this way

trainite/trainers/decoder_trainer.py
TahaZahid05: we don't use `self.scheduler` either. the point is that it should be consistent with other variables. Also, if user wants to debug related to early_stopping than they should be able to access the variable.

trainite/shared/utils.py
TahaZahid05: The line is buggy. `checkpoint_last` can never be accessed?

trainite/trainers/decoder_trainer.py
aaishwarymishra: early stopping has no attribute so I don't see the point of having it, similarly scheduler could be removed and not saved.

TahaZahid05: let's ask victor since I am not sure whether we should return or not.

TahaZahid05 (approved):
We can handle function's return value in another pr after discussion on Monday

PR #88, opened 2026-07-06
split setup_checkpointing

TahaZahid05:
### Overview
This pull request refactors the monolithic checkpoint setup function `setup_checkpointing` in `trainite/shared/utils.py` into two focused helper functions:
1. `setup_training_checkpointing` (for saving `last.pt` checkpoint on training epoch completion).
2. `setup_best_model_checkpoint` (for saving `best.pt` checkpoint on validation completion based on loss).

### DIFF
--- a/trainite/shared/utils.py
+++ b/trainite/shared/utils.py
@@ -14,6 +14,7 @@ from ignite.handlers import (
     EarlyStopping,
     create_lr_scheduler_with_warmup,
 )
+from ignite.handlers.checkpoint import CheckpointEvents
 from ignite.handlers.fbresearch_logger import FBResearchLogger
 from ignite.handlers.param_scheduler import ParamScheduler
 from ignite.handlers.tensorboard_logger import TensorboardLogger
@@ -372,3 +373,38 @@ def setup_console_logger(
         output_transform=lambda output: {"loss": output["loss"].item()},
     )
     return logger
+
+
+def setup_wandb_checkpoint_uploads(
+    trainer: Engine,
+    val_evaluator: Engine,
+    checkpointers: dict[str, Any],
+    exp_logger: Any,
+    run_name: str,
+    logger: logging.Logger,
+) -> None:
+    """Register events and upload handlers to automatically log checkpoints to W&B in real-time."""
+
+    val_evaluator.register_events(*CheckpointEvents)
+    trainer.register_events(*CheckpointEvents)
+
+    def upload_best_model_artifact(engine):
+        best_checkpoint = checkpointers.get("checkpoint_best")
+        checkpoint_path = best_checkpoint.last_checkpoint if best_checkpoint else None
+        if checkpoint_path and Path(checkpoint_path).exists():
+            logger.info(f"Uploading new best model artifact to W&B: {checkpoint_path}")
+            artifact = exp_logger.Artifact(name=f"{run_name}-model".replace("/", "-"), type="model")
+            artifact.add_file(str(checkpoint_path))
+            exp_logger.log_artifact(artifact)
+
+    def upload_last_checkpoint_artifact(engine):
+        last_checkpoint = checkpointers.get("checkpoint_last")
+        checkpoint_path = last_checkpoint.last_checkpoint if last_checkpoint else None
+        if checkpoint_path and Path(checkpoint_path).exists():
+            logger.info(f"Uploading last checkpoint artifact to W&B: {checkpoint_path}")
+            artifact = exp_logger.Artifact(name=f"{run_name}-checkpoint".replace("/", "-"), type="checkpoint")
+            artifact.add_file(str(checkpoint_path))
+            exp_logger.log_artifact(artifact)
+
+    val_evaluator.add_event_handler(CheckpointEvents.SAVED_CHECKPOINT, upload_best_model_artifact)
+    trainer.add_event_handler(CheckpointEvents.SAVED_CHECKPOINT, upload_last_checkpoint_artifact)
--- a/trainite/shared/utils.py
+++ b/trainite/shared/utils.py
@@ -357,23 +357,6 @@ def setup_experiment_tracking(
     return exp_logger
 
 
-def upload_model_to_wandb(
-    exp_logger: Any,
-    checkpointers: dict[str, Checkpoint],
-    run_name: str,
-    logger: logging.Logger,
-) -> None:
-    checkpoint = checkpointers.get("checkpoint_best")
-    checkpoint_path = checkpoint.last_checkpoint if checkpoint else None
-    if not checkpoint_path:
-        logger.warning("No checkpoint found; skipping model upload to Weights & Biases.")
-        return
-    artifact = exp_logger.Artifact(name=f"{run_name}-model".replace("/", "-"), type="model")
-    artifact.add_file(str(checkpoint_path))
-    exp_logger.log_artifact(artifact)
-    logger.info("Uploaded model checkpoint to Weights & Biases: %s", checkpoint_path)
-
-
 def setup_console_logger(
     run_dir: Path,
     engine: Engine,
--- a/trainite/shared/utils.py
+++ b/trainite/shared/utils.py
@@ -255,14 +255,28 @@ def attach_early_stopping(
     return None
 
 
-def setup_checkpointing(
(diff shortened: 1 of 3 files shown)


trainite/shared/utils.py
```diff
@@ -255,14 +255,28 @@ def attach_early_stopping(
...
+
+def setup_best_model_checkpoint(
+    engine: Engine,
+    val_evaluator: Engine,
+    to_save: dict[str, Any],
+    run_dir: Path,
+) -> Checkpoint:
     def score_function(engine_val):
         loss = engine_val.state.metrics["loss"]
         return -loss
```
aaishwarymishra: Maybe its better, to take score function as argument too, we can define one at trainer. But the thing would be where will it pass cleanely.

TahaZahid05: we can add as optional args to `setup_best_model_checkpoint`?

trainite/trainers/decoder_trainer.py
```diff
@@ -148,6 +149,36 @@ def __init__(self, config: ProjectConfig) -> None:
...
+                best_checkpoint = self.checkpointers.get("checkpoint_best")
+                checkpoint_path = best_checkpoint.last_checkpoint if best_checkpoint else None
+                if checkpoint_path and Path(checkpoint_path).exists():
+                    self.logger.info(f"Uploading new best model artifact to W&B: {checkpoint_path}")
+                    artifact = self.exp_logger.Artifact(
+                        name=f"{config.output.run_name}-model".replace("/", "-"), type="model"
+                    )
+                    artifact.add_file(str(checkpoint_path))
+                    self.exp_logger.log_artifact(artifact)
+
```
aaishwarymishra: we should only upload when the best model changes, it will slow the training, and how you make sure it uploads the bes model, the case could be the last checkpoint triggered it.

trainite/trainers/decoder_trainer.py
```diff
@@ -148,6 +149,36 @@ def __init__(self, config: ProjectConfig) -> None:
...
+                last_checkpoint = self.checkpointers.get("checkpoint_last")
+                checkpoint_path = last_checkpoint.last_checkpoint if last_checkpoint else None
+                if checkpoint_path and Path(checkpoint_path).exists():
+                    self.logger.info(f"Uploading last checkpoint artifact to W&B: {checkpoint_path}")
+                    artifact = self.exp_logger.Artifact(
+                        name=f"{config.output.run_name}-checkpoint".replace("/", "-"), type="checkpoint"
+                    )
+                    artifact.add_file(str(checkpoint_path))
+                    self.exp_logger.log_artifact(artifact)
+
```
aaishwarymishra: same here

trainite/trainers/decoder_trainer.py
aaishwarymishra, line 1: I think this is polluting the init method

trainite/shared/utils.py
aaishwarymishra: I think it should take it as required. Also I think we sholuld first merge the refactored checkpointing code, the artifact uploading can be a separate pr itself.

TahaZahid05: Why not have a default score_function and score_name and then accept if user want to override. I think its better. Both refactored, and artifact were related to checkpointing so i think it fits in one pr.

trainite/trainers/decoder_trainer.py
TahaZahid05: It does only upload when best model changes? Best and last checkpoint are scope of different engines?

trainite/trainers/decoder_trainer.py
TahaZahid05: same answer as above. they both have different scopes. and the event is only fired when a new model is written to disk

trainite/trainers/decoder_trainer.py
TahaZahid05: we can create a function instead once we are clear on the other issues with the code.

trainite/shared/utils.py
aaishwarymishra: Not sure , its cleaner to send score function themselves.

trainite/trainers/decoder_trainer.py
aaishwarymishra: hmm, my bad I got confused

trainite/trainers/decoder_trainer.py
aaishwarymishra: okay

trainite/shared/utils.py
TahaZahid05: ok

trainite/trainers/decoder_trainer.py
```diff
@@ -286,6 +254,37 @@ def _run_evaluations(self, engine: Engine) -> None:
             val_metrics["token_accuracy"],
         )
 
+    def _setup_wandb_checkpoint_uploads(self) -> None:
+        """Register events and upload handlers to automatically log checkpoints to W&B in real-time."""
+        self.val_evaluator.register_events(*CheckpointEvents)
+        self.trainer.register_events(*CheckpointEvents)
+
+        # 1. Upload best model artifact immediately when a better one is saved
+        @self.val_evaluator.on(CheckpointEvents.SAVED_CHECKPOINT)
```
aaishwarymishra: should use attach_handler method for consistency

trainite/shared/utils.py
```diff
@@ -276,17 +272,15 @@ def setup_best_model_checkpoint(
     val_evaluator: Engine,
     to_save: dict[str, Any],
     run_dir: Path,
+    score_function: Callable[[Engine], float],
```
aaishwarymishra: I think callable is enough to keep it simple for the user

trainite/trainers/decoder_trainer.py
```diff
@@ -112,24 +96,33 @@ def __init__(self, config: ProjectConfig) -> None:
...
-        attach_early_stopping(self.val_evaluator, self.engine, config.trainer.early_stopping_patience)
+        attach_early_stopping(self.val_evaluator, self.trainer, self.trainer_config.early_stopping_patience)
+
+        # Define validation scoring function for best model
+        def score_function(engine_val):
+            loss = engine_val.state.metrics["loss"]
+            return -loss
 
         # Attach checkpointing
         self.checkpointers = {
```
aaishwarymishra: its cleaner to define self.checkpointers as empty dict first, and then individually adding checkpoints

TahaZahid05:
@aaishwarymishra done.

trainite/trainers/decoder_trainer.py
```diff
@@ -234,6 +253,36 @@ def _run_evaluations(self, engine: Engine) -> None:
             val_metrics["token_accuracy"],
         )
 
+    def _setup_wandb_checkpoint_uploads(self) -> None:
```
aaishwarymishra: I think its more suited for utils.py

TahaZahid05:
@aaishwarymishra done.

trainite/shared/utils.py
```diff
@@ -389,3 +373,38 @@ def setup_console_logger(
...
+        if checkpoint_path and Path(checkpoint_path).exists():
+            logger.info(f"Uploading new best model artifact to W&B: {checkpoint_path}")
+            artifact = exp_logger.Artifact(name=f"{run_name}-model".replace("/", "-"), type="model")
+            artifact.add_file(str(checkpoint_path))
+            exp_logger.log_artifact(artifact)
+
+    def upload_last_checkpoint_artifact(engine):
+        last_checkpoint = checkpointers.get("checkpoint_last")
+        checkpoint_path = last_checkpoint.last_checkpoint if last_checkpoint else None
+        if checkpoint_path and Path(checkpoint_path).exists():
```
aaishwarymishra: isnt this check redundant if we using checkpoint saved event. Also can you check with a wandb key if it actually works

TahaZahid05: true, i was being a bit more defensive to ensure file exists even though event fires. i will change ig

aaishwarymishra (approved):

aaishwarymishra:
I would suggest running an experiment to see if artifacts gets uploaded before merging.

TahaZahid05:
@aaishwarymishra sorry didn't mention it, but I did run experiment and artifacts are getting uploaded correctly.

PR #90, opened 2026-07-06
upload and copy config.yaml as YAML file for WandB and TensorBoard

TahaZahid05:
### Overview
This pull request improves how the training configuration (`config.yaml`) is logged and stored during an experiment run:
1. **Weights & Biases (WandB)**: Uploads the physical `config.yaml` file to the run using the `exp_logger.save` method, allowing it to be viewed and downloaded under the **Files** tab of the W&B run dashboard.
2. **TensorBoard**: Copies the `config.yaml` file directly into the `tensorboard/` logging directory. This ensures the configuration is co-located with the binary event logs for easy archival and offline auditing.
3. **Cleaned up Inline Dictionary Logs**: Removed the previous behavior of stringifying the Pydantic config dictionary (`str(config_data)`) and logging it as a preformatted HTML/text block under the `"config"` key.

### DIFF
--- a/trainite/shared/utils.py
+++ b/trainite/shared/utils.py
@@ -1,7 +1,6 @@
 import importlib
 import inspect
 import logging
-import shutil
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Callable, Literal, TypeVar
@@ -304,7 +303,7 @@ def setup_experiment_tracking(
     else:
         log_dir = run_dir / "tensorboard"
         log_dir.mkdir(parents=True, exist_ok=True)
-        shutil.copy2(run_dir / "config.yaml", log_dir / "config.yaml")
+        (log_dir / "config.yaml").write_bytes((run_dir / "config.yaml").read_bytes())
         exp_logger = TensorboardLogger(log_dir=log_dir)
 
     # Log training iteration loss
--- a/trainite/shared/utils.py
+++ b/trainite/shared/utils.py
@@ -1,6 +1,7 @@
 import importlib
 import inspect
 import logging
+import shutil
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Literal, TypeVar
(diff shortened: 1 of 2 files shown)


aaishwarymishra:
there are conflicts

trainite/shared/utils.py
```diff
@@ -303,8 +304,13 @@ def setup_experiment_tracking(
 ) -> TensorboardLogger | WandBLogger:
     if backend == "wandb":
         exp_logger = WandBLogger(project=run_name, dir=str(run_dir), name=str(run_dir).split("/")[-1])
+        if run_dir and (run_dir / "config.yaml").exists():
```
aaishwarymishra: why we checking if config.yaml exists? we are not creating it anywhere

trainite/trainers/decoder_trainer.py
```diff
@@ -200,9 +200,6 @@ def _eval_step(self, engine: Engine, batch: dict[str, torch.Tensor]) -> dict[str
 
     def run(self) -> None:
         self.logger.info("starting run in %s", self.run_dir)
-        config_data = self.config.model_dump(by_alias=True, polymorphic_serialization=True)
```
aaishwarymishra, line 204: maybe a simple approach will be to basically mabe be dump config locally with simple path.write(), and if wandb is present we log it as an artifact or file.

if goinig for artifact route we can maybe make an helper as its used in model upload too, depends actually.

TahaZahid05: that is what is happening already in `decoder_trainer.py`? we dump config locally and then check wandb if present to upload the .yaml. I don't understand what you mean

trainite/shared/utils.py
TahaZahid05: my bad.

trainite/trainers/decoder_trainer.py
aaishwarymishra: where exactly I dont remember to be honest

TahaZahid05: in `__init__`:
```python
        # Create run directory for outputs
        self.run_dir = make_run_dir(config.output)
        dump_config(self.config, self.run_dir / "config.yaml")
        .
        .
        (later)
         # Attach experiment logger (TensorBoard or Weights & Biases)
        self.exp_logger = setup_experiment_tracking(
            config.logger,
            self.trainer,
            self.val_evaluator,
            self.train_evaluator,
            self.test_evaluator,
            self.optimizer,
            self.run_dir,
            ["loss", "token_accuracy"],
            bool(self.test_loader),
            config.output.run_name,
        )
```

aaishwarymishra: okay my bad, so like instead of using extra import of shutil , maybe we can use pathlib method

TahaZahid05: done.

aaishwarymishra:
can you apply pre-commit :)

TahaZahid05:
@aaishwarymishra done.

PR #91, opened 2026-07-06
extract _flatten to standalone and relocate Trainer.run and evaluations

TahaZahid05:
### Overview
This pull request refactors the `Trainer` class to improve code decoupling, class layout readability, and unit testability.


aaishwarymishra (commented):
I think test should also come up before run, and flatten accuracy maybe cana static method or be at top before trainer

aaishwarymishra (approved):

PR #99, opened 2026-07-11
Refactor dependency mocking and suppress Pydantic warnings during package initialization

TahaZahid05:
## Description

This PR resolves a critical issue where running `pip install -e .` or importing the package (e.g. via linting, type-checking, or build-backend hooks) fails with `ModuleNotFoundError: No module named 'torch'` in environments without `torch` and `pytorch-ignite` pre-installed.

It also silences verbose Pydantic warnings (`ArbitraryTypeWarning`) generated when mock modules are resolved inside Pydantic model configurations.

### DIFF
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -161,3 +161,29 @@ def test_cli_main_routing():
             ]
         )
         mock_init_project.assert_called_once()
+
+
+def test_import_without_dependencies() -> None:
+    """
+    Test that the trainite package and CLI main entrypoint can be imported
+    successfully.
+    """
+    code = (
+        "import importlib, sys\n"
+        "orig_import = importlib.import_module\n"
+        "def my_import(name, package=None):\n"
+        "    if name in ('torch', 'ignite') or (package and package.startswith(('torch', 'ignite'))):\n"
+        "        raise ImportError(f'Mocked ImportError for {name}')\n"
+        "    return orig_import(name, package)\n"
+        "importlib.import_module = my_import\n"
+        "import trainite\n"
+        "import trainite.cli.main\n"
+    )
+    result = subprocess.run(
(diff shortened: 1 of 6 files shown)


aaishwarymishra:
I think we can also modify maybe cli test which check if code runs, instead of using system installed packages, it should install them in isolated environment for testing. Currently it passes as aslong dev packages are in our system ig.

Not sure actually can you check?

TahaZahid05:
> I think we can also modify maybe cli test which check if code runs, instead of using system installed packages, it should install them in isolated environment for testing. Currently it passes as aslong dev packages are in our system ig.
> 
> Not sure actually can you check?

Yeah, as discussed we would need to change the CI. i think its best we deal this in a seperate PR as it is a change for all tests.

tests/cli_test.py
```diff
@@ -161,3 +161,29 @@ def test_cli_main_routing():
...
+        "import trainite.cli.main\n"
+    )
+    result = subprocess.run(
+        [sys.executable, "-c", code],
+        check=True,
+        timeout=10,
+        capture_output=True,
+        text=True,
+    )
+    assert "is not a Python type" not in result.stderr
```
aaishwarymishra, line 189: what does this do?

trainite/cli/mocker.py
```diff
@@ -43,5 +44,9 @@ def mock_dependencies(*packages):
             to_mock.append(pkg)
 
     if to_mock:
+        from pydantic.warnings import ArbitraryTypeWarning
```
aaishwarymishra: import can be moved to the top

tests/cli_test.py
TahaZahid05: asserts no mock-related warnings are printed for torch and ignite

PR #102, opened 2026-07-16
multi-platform tests

aaishwarymishra:
This pull request updates the GitHub Actions workflow configuration for CPU tests to improve test coverage and resource allocation. The most important changes are:

**Workflow triggers:**

* The workflow will no longer be triggered by changes to files in the `examples/**` directory.

**Test matrix and resource allocation:**

* The test matrix now includes multiple operating systems (`ubuntu-latest`, `macos-latest`, and `windows-latest`), expanding test coverage beyond just Ubuntu.
* The test job timeout has been increased from 30 to 60 minutes to accommodate potentially longer test runs across different platforms.
* The `runs-on` parameter is now set dynamically based on the operating system defined in the matrix, instead of being hardcoded to Ubuntu.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,18 @@
+import torch.utils.data
+import pytest
+
+# Store the original init method
+_original_dataloader_init = torch.utils.data.DataLoader.__init__
+
+
+def patched_dataloader_init(self, *args, **kwargs):
+    # Override num_workers to 0 to avoid process spawning overhead during tests
+    kwargs["num_workers"] = 0
+    _original_dataloader_init(self, *args, **kwargs)
+
+
+@pytest.fixture(autouse=True, scope="session")
+def force_dataloader_num_workers_zero():
+    torch.utils.data.DataLoader.__init__ = patched_dataloader_init
+    yield
+    torch.utils.data.DataLoader.__init__ = _original_dataloader_init
--- a/trainite/cli/init.py
+++ b/trainite/cli/init.py
@@ -395,6 +395,6 @@ def init_project(config: Init) -> None:
     for filename, content in templates.items():
         _write_file(resolved_project_dir / filename, content, force)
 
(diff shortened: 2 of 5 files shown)


trainite/cli/init.py
```diff
@@ -395,6 +395,6 @@ def _inject_collate(config: Any):
     for filename, content in templates.items():
         _write_file(resolved_project_dir / filename, content, force)
 
-    print("\n✔ Generated config.yaml")
+    print("\n Generated config.yaml")
```
TahaZahid05, line 398:
```suggestion
    print("\nGenerated config.yaml")
```

trainite/cli/init.py
```diff
@@ -395,6 +395,6 @@ def _inject_collate(config: Any):
...
     for filename in templates:
-        print(f"✔ Generated {filename}")
+        print(f" Generated {filename}")
```
TahaZahid05, line 400:
```suggestion
        print(f"Generated {filename}")
```

tests/conftest.py
```diff
@@ -0,0 +1,18 @@
+import torch.utils.data
+import pytest
+
+# Store the original init method
+_original_dataloader_init = torch.utils.data.DataLoader.__init__
+
+
+def patched_dataloader_init(self, *args, **kwargs):
```
TahaZahid05, line 8: what if `args` contain `num_workers`? Should set both to 0 incase someone sets `num_workers` in args to non-zero.

trainite/cli/init.py
aaishwarymishra: I think slight space looks good on cli

tests/conftest.py
aaishwarymishra: That can be the case but we have explicit keyword num_workers, in Dataloader config so it should be fine actually :)

TahaZahid05: ok. my bad.

PR #111, opened 2026-07-24
split Transformer into BasicTransformerModel and RoPETransformerModel

TahaZahid05:
## Summary

This PR separates the single `TransformerModel` into two distinct decoder-only transformer model architectures:

1. **`BasicTransformerModel`** (`basic-transformer`): Decoder-only transformer using **Absolute Positional Encoding** (`nn.Embedding(max_seq_len, hidden_size)`).
2. **`RoPETransformerModel`** (`rope-transformer`): Decoder-only transformer using **Rotary Position Embeddings (RoPE)** applied to attention queries and keys.

This gives users a clear choice between classic absolute positional encoding and relative rotary positional embeddings when scaffolding starter projects via `trainite init`.

---


trainite/models/basic_transformer.py
```diff
@@ -0,0 +1,180 @@
...
+        num_heads: int = 2,
+        feedforward_dim: int = 128,
+        dropout: float = 0.1,
+        max_seq_len: int = 128,
+        pad_token_id: int | None = None,
+    ) -> None:
+        super().__init__()
+        pad_token_id = pad_token_id if pad_token_id is not None else 0
+        self.embedding = nn.Embedding(vocab_size, hidden_size, padding_idx=pad_token_id)
+        self.pos_emb = nn.Embedding(max_seq_len, hidden_size)
```
aaishwarymishra: why using embedding instead of abs?

TahaZahid05: apologies, fixed.

aaishwarymishra (approved):

PR #113, opened 2026-07-24
Configurable tokenizer vocab

aaishwarymishra:
This pull request introduces a complete, modular string reversal example for sequence-to-sequence modeling using a Transformer architecture. It provides a full configuration-driven pipeline, including dataset generation, preprocessing, model definition, and training orchestration. The main themes are configuration management, dataset and transformation utilities, model implementation, and project entry point.

**Configuration and Project Structure:**

* Added a comprehensive `ProjectConfig` Pydantic schema in `config.py` to define all components (preprocessor, model, optimizer, scheduler, data, trainer, output, logger, seed, device) with validation and type safety.
* Introduced a YAML configuration file (`config.yaml`) that specifies all experiment parameters and links to the corresponding Python classes using `_target_` fields.

**Dataset and Preprocessing Utilities:**

* Implemented `StringReverseDataset` for generating random string reversal pairs, with support for variable sequence lengths and charset presets, and `PromptCompletionTransform` for preparing data for causal language modeling.
* Added a generic `TransformedDataset` wrapper to apply transformations to datasets in a composable way.

**Model and Collation:**

* Developed a minimal Transformer model with rotary positional embeddings, multi-head attention, and a custom collate function (`CausalLMCollateFn`) for autoregressive training.

**Project Entrypoint:**

* Added a `main.py` script to parse configuration, instantiate the trainer, and run the experiment, enabling reproducible and configurable training runs.

### DIFF
--- a/trainite/datasets/string_reverse.py
+++ b/trainite/datasets/string_reverse.py
@@ -8,7 +8,7 @@ from pydantic import BaseModel, ConfigDict
 from torch.utils.data import Dataset
 
 CHARSET_PRESETS = {
-    "@universal": string.printable,
+    "@universal": string.ascii_letters + string.digits + string.punctuation + " ",
     "@alpha": string.ascii_letters,
     "@alpha_lowercase": string.ascii_lowercase,
     "@alpha_uppercase": string.ascii_uppercase,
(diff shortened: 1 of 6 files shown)


trainite/datasets/string_reverse.py
```diff
@@ -8,7 +8,7 @@
 from torch.utils.data import Dataset
 
 CHARSET_PRESETS = {
```
TahaZahid05, line 10: I understand that we have to keep charset_preset in both `string_reverse.py` and `char_tokenizer.py` since they can be used independently but this also means redundant code will be generated if both modules are used. Maybe, we can solve this in a future PR.

TahaZahid05 (approved):

PR #121, opened 2026-08-01
make readme more focused

aaishwarymishra:
Make the old super long readme more focused and concise.

### DIFF
--- a/examples/string_reversal/README.md
+++ b/examples/string_reversal/README.md
@@ -9,21 +9,10 @@ target: 321cba
 
 The example is designed for studying how sequence length, model depth, and model width affect exact string reversal.
 
-## Run the example
-
-From this directory:
+The base tranite project which was used and modified here for the use case was generated using `trainite==0.1.0` with command
 
 ```bash
-uv sync
-uv run python main.py config.yaml
-```
-
-Training outputs are written below `outputs/<run_name>/<timestamp>/`. Each run contains its resolved configuration, console log, TensorBoard data, and model checkpoints.
-
-View TensorBoard metrics with:
-
-```bash
-uv run tensorboard --logdir outputs
+trainite init string_reversal --dataset string-reverse
 ```
 
 ## Task and data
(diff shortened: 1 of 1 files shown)


examples/string_reversal/README.md
TahaZahid05, line 1: Add the trainite command you used to generate the initial experiment folder. You can look at my counting readme for more reference.

examples/string_reversal/README.md
```diff
@@ -1,608 +1,145 @@
-# Trainite Project: string_reversal
+# String Reversal
```
TahaZahid05: Maybe you should keep this same as the blog title you use.

examples/string_reversal/README.md
```diff
@@ -1,608 +1,145 @@
...
-
-`head_dim` (which is `hidden_size // num_heads`) must be even because RoPE rotates pairs of dimensions.
-
-## Config knobs
+```bash
+uv run tensorboard --logdir outputs
+```
 
-### `hidden_size`
+## Task and data
```
TahaZahid05, line 18: This should come before `Run the example` in my opinion. We should first explain the problem and task to the reader and then how to run the code to reproduce results etc.

examples/string_reversal/README.md
aaishwarymishra: I don't think the example should be like my blog title, as its too fancy for an example like "Exploring Small Transformers with Trainite on String-Reversal"

TahaZahid05: It can be misguiding to people reading the blog whether the example in trainite is the same code you are discussing in the blog. i suggest you ask victor if current title is fine or should you keep blog title.

aaishwarymishra: @vfdev-5 thoughts?

vfdev-5: Let's copy the blog's title here. It will be consistent with Taha's example readme

TahaZahid05 (approved):

PR #127, opened 2026-08-04
Use project name when using clearml

### LINKED ISSUE #110: The default run name being generated
Currently how the run name is being generated like

```python
resolved_run_name = run_name or f"{model}__{dataset}"
```

During project run the run directory is created like

```python
def make_run_dir(output_config: Any) -> Path:
    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
    run_name = output_config.run_name
    run_dir = Path(output_config.root) / run_name / timestamp
    run_dir.mkdir(parents=True, exist_ok=True)
    return run_dir
````

So the dataset name can be `string-reverse` so the directory will be `transformer__string-reverse`, which looks weird and may fail. 

I think we could use some alternatives
1. Use generic name like run_name = "default_run" or "run1"
2. Use project name itself as the nested run name.
3. Normalize the model and the dataset name

### PR DESCRIPTION
**Pull request by aaishwarymishra**
fixes #110

### CODE REVIEW

**Comment by TahaZahid05** [comment]
@aaishwarymishra #110 also mentions concerns regarding `resolved_run_name`. Shouldn't we also update that since it still contains the weird naming.

**Comment by TahaZahid05** [comment]
@aaishwarymishra I ran the code and it still has:
```yaml
  run_name: rope-transformer__string-reverse
```

I made the experiment using `trainite init` interactive run. Can you check whether we need to normalize the run name seperately for interactive run

aaishwarymishra:
done :)

PR #130, opened 2026-08-17
Removing conftest.py

TahaZahid05:
This PR aims to remove conftest.py since it forces `num_workers` to 0. This seems to be incorrect and not required.


aaishwarymishra:
? reason

TahaZahid05:
@aaishwarymishra As written in the description, there is no need to set `num_workers` to 0. The code written in file is not necessary.

aaishwarymishra:
The reason it was added because macos and windows python uses spawn() which spins up entire new python process https://docs.pytorch.org/docs/main/data.html#platform-specific-behaviors

This can make an training test currently integration test which can finish in 5 min in Linux take up to 10-15 minutes due to num_workers being 2.

PR #133, opened 2026-08-19
Remove redundant model configs

### LINKED ISSUE #132: Remove unnecessary model configs from trainite/config
Currently config only file which is being generated is `base.py`, all the other configs are only treated as read-only meaning the pydantic models that have model-config like

```python
 model_config = ConfigDict(extra="allow", validate_assignment=True)
```

Are not necessary and can be removed, one thing to consider few pydantic models do require the `model_config` and must be verified before removing an item.

yansh07:
## Summary

- Remove redundant `model_config` declarations from specialized config classes.
- Rely on the configuration inherited from the corresponding base config classes.
- Remove now-unused `ConfigDict` imports.

## Testing

- `uv run pytest`
- 105 passed, 5 skipped

Closes #132

### DIFF
--- a/trainite/config/datasets.py
+++ b/trainite/config/datasets.py
@@ -1,10 +1,9 @@
 from typing import Literal, Self
-from pydantic import Field, ConfigDict, model_validator
+from pydantic import Field, model_validator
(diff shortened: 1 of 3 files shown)


aaishwarymishra:
Thanks!

PR #134, opened 2026-08-22
Import self from typing_extensions

TahaZahid05:
This PR imports `Self` from `typing_extensions` since Python 3.10 doesn't support `from typing import Self`


aaishwarymishra:
I think we are supporting python3.11 and onwards

PR #139, opened 2026-08-30
fix(cli): handle cancelled model selection

### LINKED ISSUE #137: Fix Cancel Handling In Interactive Mode
Currently, when the user presses terminates `trainite init` on model selection i.e presses Ctrl+C when selecting model, it prints a misleading error:

```bash
trainite init
? Project directory:  Directory to create the starter project in
 my-cool-experiment
? Model(s): Select starter model template(s) to include (use space to select)

Cancelled by user

At least one model must be selected.
```

It prints `At least one model must be selected` which is misleading as the command was terminated rather than failed due to an error. This needs correct error handling so it doesn't print a misleading error.

cananoo:
## Summary

- exit cleanly when the interactive model checkbox is cancelled
- keep the existing validation error for an empty model selection
- add a regression test for the cancellation path

Fixes #137

## Testing

- `uv run pytest tests/cli_test.py -k interactive_model_selection_cancel_exits_cleanly -q`
- `PYTHONUTF8=1 uv run pytest -q`
- `uv run pre-commit run --all-files`

### DIFF
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -245,6 +245,20 @@ def test_duplicate_models_raises_error():
         Init(model=("rope-transformer", "rope-transformer"))
 
 
+def test_interactive_model_selection_cancel_exits_cleanly():
+    from unittest import mock
+
+    from trainite.cli.init import _prompt_multi_choice
+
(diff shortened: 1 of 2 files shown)


TahaZahid05:
Closed by #151

PR #146, opened 2026-08-30
Add tests for the component registry

### LINKED ISSUE #140: Add Tests For The Component Registry
## Description

The component registry defines the models, datasets, trainers, and preprocessors available to the CLI, but its static entries are not directly tested. A broken path or target can therefore be discovered only when a project is generated.

This is a focused part of #31.

## Task

Add a small test module for `trainite/config/registry.py` that verifies:

- The expected component names are present in each registry group.
- Every `implementation_path` exists.
- Every configured `readme_template_path` exists when provided.
- Every `config_cls_path` and `dataset_config_cls_path` can be resolved.
- Every configured `collate_fn_target` can be resolved.

## Acceptance criteria

- Only tests are changed unless a test exposes a genuine registry bug.
- Tests use the existing registry objects and `get_target`; do not duplicate registry data or introduce new abstractions.
- `uv run pytest` passes.

pokemon225:
The registry defines the models, datasets, trainers and preprocessors exposed by the CLI, but its static entries were not covered by tests, so a broken path or target only surfaced when a project was generated.

Add tests/config/registry_test.py, which asserts that each registry group contains the expected component names, that every implementation_path and configured readme_template_path exists on disk, and that every config_cls_path, dataset_config_cls_path and collate_fn_target resolves through get_target.

Tests only; the registry itself is unchanged.

Closes #140

### DIFF
--- a/tests/config/registry_test.py
+++ b/tests/config/registry_test.py
@@ -1,0 +1,87 @@
+from pathlib import Path
+
+import pytest
+
+import trainite
+from trainite.config.registry import (
+    DATASET_SPECS,
+    MODEL_SPECS,
+    PREPROCESSOR_SPECS,
+    REGISTRY,
+    TRAINER_SPECS,
+    DatasetSpec,
+    ModelSpec,
+)
+from trainite.shared.utils import get_target
+
+REPO_ROOT = Path(trainite.__file__).resolve().parent.parent
+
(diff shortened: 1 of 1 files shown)


aaishwarymishra:
Hi, this Issue was for an event, we still appreciate your pr :)

PR #147, opened 2026-08-30
Remove obsolete vocab_size resolution tests

### LINKED ISSUE #138: Clean Up Obsolete Tests
Some tests inside the repository are currently skipped as we transitioned the codebase towards other design choices. We can remove these tests if they are no longer relevant.

pokemon225:
Three tests in tests/trainers/decoder_trainer_test.py were skipped with the reason "Obsolete after decoupling tokenizer from model and dataset vocab_size resolution": test_decoder_trainer_auto_vocab_size, test_decoder_trainer_vocab_size_mismatch and
test_decoder_trainer_vocab_size_missing. They assert behaviour the trainer no longer implements, so remove them.

SimpleDatasetNoVocab was only referenced by the last of these, so it is removed as well rather than left as dead code.

The two tests skipped with "validation is required" are left in place: that reason describes a test that needs updating rather than one made obsolete by a design change, so it seemed out of scope here. Happy to remove those too if they are also considered obsolete.

Closes #138

### DIFF
--- a/tests/trainers/decoder_trainer_test.py
+++ b/tests/trainers/decoder_trainer_test.py
@@ -74,21 +74,6 @@ class SimpleDataset(torch.utils.data.Dataset):
         }
 
 
-class SimpleDatasetNoVocab(torch.utils.data.Dataset):
-    def __init__(self, size=16, seq_len=4, **kwargs):
-        self.size = size
-        self.seq_len = seq_len
-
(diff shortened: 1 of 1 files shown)


TahaZahid05:
Thank you for your PR but this is issue for a specific event! Kindly look at Issues not tagged with CAT Reloaded tag.

PR #151, opened 2026-09-02
fix(cli): distinguish Ctrl+C from empty selection in model prompt

### LINKED ISSUE #137: Fix Cancel Handling In Interactive Mode

AmrTamer217:
## Description

Fixes the misleading error message shown when a user cancels `trainite init` during model selection.

Previously, `_prompt_multi_choice` treated two different outcomes from the questionary checkbox prompt identically:

- The user cancelling via Ctrl+C/EOF (`ask()` returns `None`)

- The user submitting the prompt with nothing checked (`ask()` returns `[]`)

Both cases raised `SystemExit("At least one model must be selected.")`, so cancelling the command printed a validation error even though nothing was actually invalid — the user just quit.

`_prompt_multi_choice` now distinguishes the two cases:

- Cancelled (`None`) → `SystemExit(0)`, with no message printed.

- Submitted empty (`[]`) → `SystemExit("At least one model must be selected.")`, unchanged from before.

## Testing

Added two tests in `tests/cli_test.py` that exercise the fix through the real CLI entry point (`main(argv=["init"])`), mocking `questionary.checkbox` to simulate each outcome:

- `test_interactive_init_ctrl_c_on_model_selection`: simulates Ctrl+C (`ask()` returns `None`), asserts the process exits with code `0` and that the misleading message does not appear in stdout/stderr. Also asserts the checkbox prompt was called exactly once, so the exit can't be misattributed to a different prompt.

- `test_interactive_init_empty_model_selection_shows_error`: simulates submitting with nothing checked (`ask()` returns `[]`), asserts the validation message is still raised as before.

## Checklist

- [x] `uv run pytest` passes

- [x] `uv run pre-commit run --all-files` passes

Closes #137


trainite/cli/init.py
```diff
@@ -79,7 +79,9 @@ def _prompt_multi_choice(
         choices=formatted_choices,
         instruction=instruction,
     ).ask()
-    if result is None or len(result) == 0:
+    if result is None:
+        raise SystemExit(0)
+    elif len(result) == 0:
         raise SystemExit("At least one model must be selected.")
```
vfdev-5: @aaishwarymishra actually why the error message speaks about "model", as the method is rather generic, the error message should be configured, I think, e.g. an argument of the function.

aaishwarymishra:
```suggestion
        raise SystemExit("At least one option must be selected.")
```

tests/cli_test.py
aaishwarymishra, line 1: For the tests, I am not big fan of mock patch approach, I was thinking on removing some other tests that use them or simplify them, for this case the fix is trivial and i don't think we need dedicated tests for them :)

AmrTamer217:
Tests are removed as requested

aaishwarymishra:
looks good to me, thanks!

PR #157, opened 2026-09-02
test: validate component registry

### LINKED ISSUE #140: Add Tests For The Component Registry

Mohammed-Adel1:
## Summary

* Add tests for component registry entries to ensure the registered components are correctly defined and accessible.
* Add validation for implementation and README template paths to ensure the referenced files exist and are correctly configured.
* Add validation for config, dataset config, and collate function targets using `get_target`.
* Improve test coverage around component registration and target resolution.

## Testing

* `uv run pytest` — **123 passed, 5 skipped**
* `uv run pre-commit run --all-files` — **passed**

Fixes #140

### DIFF
--- a/tests/config/registry_test.py
+++ b/tests/config/registry_test.py
@@ -1,0 +1,52 @@
+from trainite.config.registry import REGISTRY
+from trainite.shared.utils import get_target
+
+
+def test_model_registry_names():
+    assert "basic-transformer" in REGISTRY["models"]
+    assert "rope-transformer" in REGISTRY["models"]
+
+
+def test_dataset_registry_names():
+    assert "string-reverse" in REGISTRY["datasets"]
+    assert "counting" in REGISTRY["datasets"]
+    assert "hugging-face" in REGISTRY["datasets"]
+
+
+def test_trainer_registry_names():
+    assert "decoder-trainer" in REGISTRY["trainers"]
+
+
+def test_preprocessor_registry_names():
+    assert "char" in REGISTRY["preprocessors"]
+    assert "gpt2" in REGISTRY["preprocessors"]
+
+
+def test_implementation_paths_exist():
+    for specs in REGISTRY.values():
+        for spec in specs.values():
+            assert spec.implementation_path.is_file()
+
+
+def test_readme_template_paths_exist():
+    for specs in REGISTRY.values():
+        for spec in specs.values():
+            if spec.readme_template_path is not None:
+                assert spec.readme_template_path.is_file()
+
+
+def test_config_cls_paths_resolve():
+    for specs in REGISTRY.values():
+        for spec in specs.values():
+            get_target(spec.config_cls_path)
+
+
+def test_dataset_config_cls_paths_resolve():
+    for spec in REGISTRY["datasets"].values():
+        get_target(spec.dataset_config_cls_path)
+
+
+def test_collate_fn_targets_resolve():
+    for spec in REGISTRY["models"].values():
+        if spec.collate_fn_target is not None:
+            get_target(spec.collate_fn_target)
--- a/tests/config/registry_test.py
+++ b/tests/config/registry_test.py
@@ -1,52 +1,36 @@
-from trainite.config.registry import REGISTRY
+from pathlib import Path
+
+import pytest
+
+from trainite.config.registry import REGISTRY, DatasetSpec, ModelSpec
 from trainite.shared.utils import get_target
 
-
-def test_model_registry_names():
-    assert "basic-transformer" in REGISTRY["models"]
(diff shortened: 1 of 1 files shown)


tests/config/registry_test.py
```diff
@@ -0,0 +1,52 @@
+from trainite.config.registry import REGISTRY
+from trainite.shared.utils import get_target
+
+
+def test_model_registry_names():
+    assert "basic-transformer" in REGISTRY["models"]
+    assert "rope-transformer" in REGISTRY["models"]
```
aaishwarymishra: I dont like these types of value checks as they are static and needs to be updated every time. You can remove them

tests/config/registry_test.py
```diff
@@ -0,0 +1,52 @@
...
+    for specs in REGISTRY.values():
+        for spec in specs.values():
+            if spec.readme_template_path is not None:
+                assert spec.readme_template_path.is_file()
+
+
+def test_config_cls_paths_resolve():
+    for specs in REGISTRY.values():
+        for spec in specs.values():
+            get_target(spec.config_cls_path)
```
aaishwarymishra: These can be combined in a single test instead of looping multiple times

tests/config/registry_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_trainer_registry_names():
+    assert "decoder-trainer" in REGISTRY["trainers"]
+
+
+def test_preprocessor_registry_names():
+    assert "char" in REGISTRY["preprocessors"]
+    assert "gpt2" in REGISTRY["preprocessors"]
+
```
aaishwarymishra: A simpler test would be to check, if keys, models, dataset  keys etc exists in the registery instead of individual components.

Mohammed-Adel1: ok , I'll review the changes and inform you .

Mohammed-Adel1:
@aaishwarymishra 
Thanks for the feedback! I've updated the tests to avoid hard-coding individual component names and now only validate the registry groups. I also consolidated the component validation into a parametrized test using the existing REGISTRY and get_target.

tests/config/registry_test.py
```diff
@@ -1,52 +1,36 @@
...
+PROJECT_ROOT = Path(__file__).resolve().parents[2]
 
-def test_preprocessor_registry_names():
-    assert "char" in REGISTRY["preprocessors"]
-    assert "gpt2" in REGISTRY["preprocessors"]
 
+def _registered_specs():
+    for group_name, specs in REGISTRY.items():
+        for component_name, spec in specs.items():
+            yield pytest.param(spec, id=f"{group_name}/{component_name}")
```
aaishwarymishra, line 14: why we need id?

Mohammed-Adel1: Without `id=`, pytest can't build a readable label from a `ComponentSpec` (it's a plain Pydantic object, not a string/number), so it falls back to positional names like `test_registry_spec_resolves[spec0]`, `spec1`, ..., `spec7`. I checked this directly.

With `id=f"{group_name}/{component_name}"`, pytest instead produces names like `test_registry_spec_resolves[preprocessors/gpt2]` and `test_registry_spec_resolves[datasets/hugging-face]`.

This is useful here because when a component breaks, the test name itself identifies the failing component without needing to inspect the traceback. Since the purpose of this test is to catch broken registry entries before project generation, a component-specific test name is more useful in CI than something like `spec4 FAILED`.

I'm happy to remove it if you'd prefer to keep the test minimal. It's not required for correctness; it's mainly for making test failures easier to identify.

aaishwarymishra:
Its good to resolve the conversations which are solved

aaishwarymishra:
This contribution is for cat-reloaded right?

aaishwarymishra (approved):

Mohammed-Adel1:
> This contribution is for cat-reloaded right?

Yes, it is .

aaishwarymishra:
Thanks!

aaishwarymishra:
My bad forgot to merge :(

PR #159, opened 2026-09-02
Remove unnecessary test __init__.py files

### LINKED ISSUE #158: Remove unnecessary `__init__.py` files from `tests/`
Pytest does not require tests folder to have `__init__.py` files. Therefore, we can delete any `__init__.py` files in `tests/` or any of its subfolders to align with pytest best practices.

Mohammed-Adel1:
## Description

This PR removes unnecessary `__init__.py` files from the `tests/` directory and its subdirectories.

### Changes

* Removed `__init__.py` files that are not required by pytest from the `tests/` directory.
* No test logic or production code was modified.
* Kept the existing test structure unchanged apart from removing the unnecessary package initialization files.

Closes #158

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,22 @@
+## Summary
+
+<!-- What does this PR change, and why? -->
+
+## AI usage
+
+<!-- Select one option and provide details when AI was used. -->
+
+- [ ] No AI tools were used for this contribution
+- [ ] AI tools were used to assist this contribution
+
+**Tool(s) and usage:**
+
+<!-- Example: GitHub Copilot for code completion; ChatGPT for test suggestions. Write "N/A" if no AI was used. -->
+
+## Testing
+
+<!-- List the checks you ran, or explain why testing was not applicable. -->
+
+- [ ] Tests added or updated where needed
+- [ ] Existing tests and checks pass
+- [ ] Documentation updated if needed
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,16 +48,23 @@ Tests are organized by component:
 
 ```
 tests/
-├── cli_test.py                     # CLI and code generation
+├── cli_test.py                     # Core CLI and code generation tests
+├── cli_sky_test.py                 # Sky integration CLI tests
+├── conftest.py                     # Shared pytest fixtures
 ├── utils_test.py                   # Config loading and instantiation
 ├── config/
-│   └── base_test.py                # Config validation
+│   └── base_test.py                # Config validation tests
 ├── datasets/
-│   └── string_reverse_test.py      # Dataset and tokenizer
+│   ├── counting_test.py            # Counting dataset tests
+│   ├── hugging_face_test.py        # Hugging Face dataset integration
+│   ├── string_reverse_test.py      # String reverse dataset tests
+│   └── transformed_test.py         # Transformed dataset wrappers
 ├── models/
 │   └── transformer_test.py         # Model forward pass and generation
+├── preprocessors/
+│   └── char_test.py                # Character tokenizer tests
 └── trainers/
-    └── decoder_trainer_test.py      # Training loop integration
+    └── decoder_trainer_test.py     # Training loop integration
 ```
 
 ---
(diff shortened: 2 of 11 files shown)


TahaZahid05 (approved):
Thanks!

TahaZahid05:
Thanks for the PR! Unfortunately, the issue has already been assigned to someone else, so we'll go ahead with their PR. Sorry!

TahaZahid05 (approved):
Thanks!

TahaZahid05:
@Mohammed-Adel1 tests are failing. Can you debug why is that so and fix?

Mohammed-Adel1:
> @Mohammed-Adel1 tests are failing. Can you debug why is that so and fix?

Ok , I will review and fix it .

TahaZahid05:
@Mohammed-Adel1 hi! are you facing any issues? Let me know if you need help.

Mohammed-Adel1:
> @Mohammed-Adel1 hi! are you facing any issues? Let me know if you need help.

Sorry for the late :)

I tested this by removing both tests/__init__.py and tests/trainers/__init__.py and running the full test suite.
This caused import errors in decoder_trainer_test.py, 

ImportError: Could not locate target
'tests.trainers.decoder_trainer_test.DummyTokenizer':
No module named 'tests.trainers'

The affected tests use dotted targets under tests.trainers..., which are resolved through get_target.
After restoring the two __init__.py files, those import errors disappeared and the affected tests passed.

So it looks like tests/trainers/__init__.py is currently required by the existing tests. Would you prefer that I keep it, or update the affected test targets to allow removing it?

TahaZahid05:
@Mohammed-Adel1 we want to debug why the tests are failing and update them if possible such that they don't rely on __init__.py

esthermichelle:
Hi @Mohammed-Adel1 and @TahaZahid05,
I looked into Issue #158 and the failing tests in PR #159 to understand why removing the `__init__.py` files causes the import errors
The issue seems to come from the hardcoded targets in `tests/trainers/decoder_trainer_test.py`, such as:

```python
"tests.trainers.decoder_trainer_test.DummyTokenizer"
```

These are resolved through `get_target()` using `importlib.import_module()`. Once the `__init__.py` files are removed, `tests.trainers` is no longer available as the expected package path.
I tried changing these targets to use the current module instead, for example:

```python
f"{__name__}.DummyTokenizer"
```

This allows the tests to resolve the objects from the module pytest is actually running, without relying on `__init__.py`
I removed both `tests/__init__.py` and `tests/trainers/__init__.py` and verified the changes:

- `uv run pytest tests/trainers/decoder_trainer_test.py` → **passed**
   
- `uv run pytest` → **passed**
    
- `uv run pre-commit run --all-files` → **all checks passed**
    

I already implemented and committed the changes here:
Branch: [https://github.com/esthermichelle/trainite/tree/fix/remove-test-init-files](https://github.com/esthermichelle/trainite/tree/fix/remove-test-init-files)
Commit: `86b435a`
If you have time, could you try this approach? If you don't have time to test it, I can open a PR with this implementation for reviewing

vfdev-5:
I think the change `"tests.trainers.decoder_trainer_test.DummyTokenizer"
` to `f"{__name__}.DummyTokenizer"` may hurt the readability of the code.

esthermichelle:
Thanks for pointing that out, I agree that using `f"{__name__}.Symbol"` throughout the tests isn't the most readable approach
I found a cleaner way to keep the tests independent of `__init__.py`. Instead of passing the target as a manually constructed string, I updated the local `cc()` helper to accept the actual class or function and generate the import path internally using its `__module__` and `__qualname__`.
So instead of:

```python
cc(f"{__name__}.SimpleModel")
```

we can simply write:

```python
cc(SimpleModel)
```

The helper converts `SimpleModel` to its dotted import path internally before passing it to `MockComponent`. I used the same approach for the direct `collate_fn_target` assignments through a small `target_path()` helper
This keeps the test code cleaner and avoids relying on `tests/__init__.py` or `tests/trainers/__init__.py`
I think this keeps the tests more readable while still avoiding the `__init__.py` dependency, Would you consider this approach preferable?

TahaZahid05:
Sounds good to me @esthermichelle. Can you open a PR that builds on top of `Mohammed-Adel1:fix/remove-test-init-files` branch?

PR #161, opened 2026-09-02
fix(tokenizer): preserve special tokens during truncation

### LINKED ISSUE #143: Preserve Special Tokens When CharTokenizer Truncates Input
## Description

When `CharTokenizer.__call__` is used with `add_special_tokens=True`, `truncation=True`, and `max_length`, it currently builds `[BOS] + tokens + [EOS]` and then keeps the first `max_length` IDs. Long inputs therefore lose the EOS token.

Example:

```python
tokenizer = CharTokenizer()
result = tokenizer("abcdef", truncation=True, max_length=5)
```

The result should fit within five tokens while retaining both BOS and EOS.

## Task

Update truncation in `trainite/preprocessors/char_tokenizer.py` so that:

- Output never exceeds `max_length`.
- BOS and EOS are retained when `add_special_tokens=True` and `max_length` has room for both.
- Behavior without special tokens remains unchanged.
- The attention mask still matches the returned IDs.

Add focused tests in `tests/preprocessors/char_test.py`.

## Acceptance criteria

- Add tests for truncation with and without special tokens.
- Define sensible behavior for `max_length` values smaller than two and test it.
- Do not add dependencies or tokenizer abstractions.
- `uv run pytest` passes.

BaselHamdy1:
### Fixes #143
This PR fixes issue #143 where `CharTokenizer` lost the `<EOS>` token when truncating text with `add_special_tokens=True`. 

### Changes Made
- Updated truncation logic in `CharTokenizer.__call__` to reserve 2 slots for `<BOS>` and `<EOS>` special tokens before slicing sequence IDs to `max_length`.
- Handled edge cases where `max_length < 2`.
- Ensured `attention_mask` matches the final output length.
- Updated and expanded unit tests in `tests/preprocessors/char_test.py` to cover truncation behavior with and without special tokens as well as edge cases.

### DIFF
--- a/trainite/preprocessors/char_tokenizer.py
+++ b/trainite/preprocessors/char_tokenizer.py
@@ -120,13 +120,17 @@ class CharTokenizer:
 
         for t in texts:
             ids = self.encode(t)
+            if truncation and max_length is not None:
+                num_special = 2 if add_special_tokens else 0
+                if max_length < num_special:
+                    ids = []
+                else:
+                    ids = ids[: max_length - num_special]
+
             if add_special_tokens:
                 prefix = [self.bos_token_id]
                 suffix = [self.eos_token_id]
                 ids = prefix + ids + suffix
-
-            if truncation and max_length is not None:
-                ids = ids[:max_length]
 
             batch_input_ids.append(ids)
             batch_attention_mask.append([1] * len(ids))
@@ -143,9 +147,6 @@ class CharTokenizer:
                     pad_len = target_len - current_len
                     batch_input_ids[i] = [self.pad_token_id] * pad_len + batch_input_ids[i]
                     batch_attention_mask[i] = [0] * pad_len + batch_attention_mask[i]
-                elif current_len > target_len:
-                    batch_input_ids[i] = batch_input_ids[i][-target_len:]
-                    batch_attention_mask[i] = batch_attention_mask[i][-target_len:]
 
         if not is_batched:
             out = {
--- a/trainite/preprocessors/char_tokenizer.py
+++ b/trainite/preprocessors/char_tokenizer.py
@@ -122,17 +122,17 @@ class CharTokenizer:
             ids = self.encode(t)
             if truncation and max_length is not None:
                 num_special = 2 if add_special_tokens else 0
-                keep = max(max_length - num_special, 0)
+                if max_length < num_special or max_length <= 0:
+                    raise ValueError(
+                        f"max_length ({max_length}) must be at least {num_special if add_special_tokens else 1} "
+                        f"when truncation=True and add_special_tokens={add_special_tokens}."
+                    )
+
+                keep = max_length - num_special
                 ids = ids[:keep]
 
-                if add_special_tokens:
-                    if max_length >= 1:
-                        ids = ids + [self.eos_token_id]
-                    if max_length >= 2:
-                        ids = [self.bos_token_id] + ids
-            else:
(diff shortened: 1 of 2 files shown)


trainite/preprocessors/char_tokenizer.py
aaishwarymishra, line 1: I think the bug could have been fixed with few lines of code i dont know why we needed this much refactoring

BaselHamdy1: so the truncation with the special tokens already was inside the code in padding == "max_length"

in other lines it was ids[:max_length] after adding BOS,EOS which chops off EOS because its the last element 

so if i fixed only 2-3 lines it would have fixed it but other bugs would have appeared 
so what i did was that i deleted the ids[:max_length] logic and kept the correct one

aaishwarymishra: In the padding == "max_length", we explicitly check for the max sequence length if max_sequence is not provided, yeah I think the elif case for truncating in the padding is wrong and can be removed so padding never truncates anything, can you maybe look at how hugging face handles this? do they truncate items in padding ?

BaselHamdy1: I simplified the code and brought back the original approachnPadding now only pads shorter sequences without touching truncationand truncation explicitly reserves space for BOS,EOS beforehand

trainite/preprocessors/char_tokenizer.py
```diff
@@ -120,14 +120,18 @@ def __call__(
 
         for t in texts:
             ids = self.encode(t)
+            if truncation and max_length is not None:
+                num_special = 2 if add_special_tokens else 0
+                if max_length < num_special:
+                    ids = []
+                else:
+                    ids = ids[: max_length - num_special]
+
```
aaishwarymishra, line 133: I dont think this is correct

aaishwarymishra:
just to confirm your contribution is for cat-reloaded?

trainite/preprocessors/char_tokenizer.py
aaishwarymishra: There are basically 2 scenario, in case special_tokens is true, then we have to reserve some positions for the special tokens, incase its false we can simply slice to max_length

BaselHamdy1: I've cleaned up the truncation logic so it handles both scenarios

trainite/preprocessors/char_tokenizer.py
```diff
@@ -120,13 +120,19 @@ def __call__(
...
+                ids = ids[:keep]
+
+                if add_special_tokens:
+                    if max_length >= 1:
+                        ids = ids + [self.eos_token_id]
+                    if max_length >= 2:
+                        ids = [self.bos_token_id] + ids
+            else:
+                if add_special_tokens:
+                    ids = [self.bos_token_id] + ids + [self.eos_token_id]
```
aaishwarymishra: This block is weird, why we checking max_length>=1 and after with 2 ?

And if keeps lengths = 0, we should throw error.

BaselHamdy1: i was trying to partially fit special tokens for max_length < 2

i modified  the logic to validate max_length upfront and raise a ValueError if it's smaller than the required special token count or <= 0

aaishwarymishra (approved):

aaishwarymishra:
thanks!

Issue #128, opened 2026-08-10
Making ci tests self contained


aaishwarymishra:
# Updated PR

Isolate and reduce generated-project integration tests

### Problem

Generated projects are currently run with the test suite’s Python environment. 
The tests also train every model/dataset combination for one epoch, making CI unnecessarily expensive.

### Proposed solution

- Keep lightweight generation checks for all combinations.
- Run one representative end-to-end smoke test in an isolated `uv` environment.
- Install only dependencies declared by the generated project.
- Limit training to one batch or the smallest practical workload.
- Rely on the shared `uv` package cache to avoid repeated downloads.

### Acceptance criteria

- [x] Generated code runs in an isolated environment.
- [x] Missing generated dependencies cause the test to fail.
- [x] One representative project completes a minimal training run.
- [x] All combinations remain covered by lightweight generation tests.

TahaZahid05:
I think while there are obvious downsides of the tests taking more time, it is still better to make the test accurate so we don't merge broken code. Testing model training with the correct dependencies is critical so that Trainite works as intended.

aaishwarymishra:
@vfdev-5 thoughts?

aniketpandey05:
@aaishwarymishra @TahaZahid05 can i pick this up?

aaishwarymishra:
Hi, sorry @aniketpandey05  this is for an event :)

saraelsayed13:
@aaishwarymishra I’d like to work on this issue as part of the Catreloaded-2026 event

vfdev-5:
@aaishwarymishra can you update the description to make it clear about what to do for this issue. 
As it was discussed previously in our meetings, we do not need now slow integration tests running on each PR, instead we can run them nightly. If we do that we should propose what kind of tests we should run on GHA (cpu only)

saraelsayed13:
@aaishwarymishra @vfdev-5 PR #166 is ready for review and implements the split discussed here.

* PR/push CI now runs only the lightweight generation checks (all combinations).
* The training-based integration test runs in an isolated `uv` env (CPU-only torch, single representative combo) and has been moved to a new nightly workflow instead of running on every PR.
* Missing generated-project dependencies now fail the test directly.

Happy to adjust anything needed.

aaishwarymishra:
completed

PR #166, opened 2026-09-03
test: isolate generated-project integration tests

saraelsayed13:
## Summary
This PR makes the generated-project integration test self-contained and removes it from the PR/push CI path to keep CI lightweight while preserving regular integration coverage.

## What changed
- The integration test now runs in an isolated `uv` environment using only the generated project's declared dependencies, so missing dependencies fail directly.
- Reduced the training test to one representative combination (`rope-transformer` + `string-reverse`) with a minimal workload.
- Uses CPU-only PyTorch to avoid unnecessary CUDA/GPU downloads in CI.
- Added the `integration` pytest marker and excluded integration tests from regular PR/push CI.
- Added a nightly workflow to run the integration test regularly.
- Lightweight generation tests still cover all model/dataset combinations on every PR.

Relates to #128

## AI usage

**Tool(s) and usage:**
Claude was used to help understand the existing test structure and debug the isolated `uv`-based integration test setup.

## Testing

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -45,4 +45,5 @@ jobs:
         run: uv sync --locked --dev
 
       - name: Run tests
-        run: uv run pytest
+        run: uv run pytest -m "not integration"
+        
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -45,5 +45,4 @@ jobs:
         run: uv sync --locked --dev
 
       - name: Run tests
-        run: uv run pytest -m "not integration"
-        
+        run: uv run pytest
--- a/.github/workflows/nightly-integration.yml
+++ b/.github/workflows/nightly-integration.yml
@@ -1,0 +1,24 @@
+name: nightly integration tests
+
+on:
+  schedule:
+    - cron: "0 3 * * *"
+  workflow_dispatch:
+
+jobs:
+  integration:
+    name: Integration test (CPU)
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+      - name: Install uv and Python
+        uses: astral-sh/setup-uv@v9.0.0
+        with:
+          enable-cache: true
+          python-version: "3.11"
+      - name: Sync dependencies
+        run: uv sync --locked --dev
+      - name: Run integration tests
+        run: uv run pytest -m integration
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -168,7 +168,7 @@ exclude = '''
 )
 '''
 [tool.pytest.ini_options]
-addopts = "--color=yes"
+addopts = "--color=yes -m 'not integration'"
 markers = [
     "integration: slow tests that build an isolated uv env and train a real step",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -167,4 +167,8 @@ exclude = '''
   | foo.py
 )
 '''
+[tool.pytest.ini_options]
 addopts = "--color=yes"
+markers = [
+    "integration: slow tests that build an isolated uv env and train a real step",
+]
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -149,21 +149,7 @@ def test_generated_project_is_runnable() -> None:
         with open(config_path, "w") as f:
             yaml.safe_dump(config, f)
 
-        # 3. Configure CPU-only PyTorch for the generated project
-        pyproject_path = project_dir / "pyproject.toml"
-        with open(pyproject_path, "a") as f:
-            f.write(
-                """
-[[tool.uv.index]]
-name = "pytorch-cpu"
-url = "https://download.pytorch.org/whl/cpu"
-explicit = true
-
-[tool.uv.sources]
-torch = { index = "pytorch-cpu" }
-"""
-            )
-        # 4. Install ONLY the generated project's own declared dependencies
+        # 3. Install ONLY the generated project's own declared dependencies
         try:
             subprocess.run(
                 ["uv", "sync"],
@@ -182,7 +168,7 @@ torch = { index = "pytorch-cpu" }
         except subprocess.TimeoutExpired:
             pytest.fail(f"uv sync timed out for generated project ({model} + {dataset})")
 
-        # 5. Run the generated main.py
+        # 4. Run the generated main.py
         try:
             subprocess.run(
                 ["uv", "run", "python", "main.py", "config.yaml"],
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -102,19 +102,13 @@ def test_init_generates_valid_project(models: list[str], dataset: str, trainer: 
         assert generated_config["data"]["transform"]["_target_"].startswith("dataset_impl.")
 
 
-@pytest.mark.parametrize(
-    "model,dataset,trainer",
-    [
-        ("basic-transformer", "string-reverse", "decoder-trainer"),
-        ("rope-transformer", "string-reverse", "decoder-trainer"),
-        ("basic-transformer", "counting", "decoder-trainer"),
-        ("rope-transformer", "counting", "decoder-trainer"),
-    ],
-)
-def test_generated_project_is_runnable(model: str, dataset: str, trainer: str) -> None:
+@pytest.mark.integration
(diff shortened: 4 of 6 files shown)


aaishwarymishra:
why did you remove AI policy part from the template :)

saraelsayed13:
I accidentally removed that section while editing. Added it back with the AI usage details.

tests/cli_test.py
```diff
@@ -147,20 +141,51 @@ def test_generated_project_is_runnable(model: str, dataset: str, trainer: str) -
...
+            f.write(
+                """
+[[tool.uv.index]]
+name = "pytorch-cpu"
+url = "https://download.pytorch.org/whl/cpu"
+explicit = true
+
+[tool.uv.sources]
+torch = { index = "pytorch-cpu" }
+"""
```
aaishwarymishra: whats this used for?

.github/workflows/cpu-tests.yml
```diff
@@ -45,4 +45,5 @@ jobs:
         run: uv sync --locked --dev
 
       - name: Run tests
-        run: uv run pytest
+        run: uv run pytest -m "not integration"
```
aaishwarymishra: can you explain what this does?

tests/cli_test.py
saraelsayed13: - I noticed `uv sync` was pulling in the full CUDA build of torch (~1.5GB with triton, cublas, etc.), which made the isolated env setup really slow on a CPU-only runner. I even bumped the timeout to 600s and it still wasn't enough the first time , so instead of just waiting longer, I forced CPU-only torch here, which fixed it properly.

- Only affects the temp project generated inside this test , doesn't touch the main `pyproject.toml` or `trainite init` for regular users.

.github/workflows/cpu-tests.yml
saraelsayed13: 1-Skips tests marked `@pytest.mark.integration` on regular PR/push 2-CI. `test_generated_project_is_runnable` is now marked as `integration` since it builds an isolated env and actually trains a model , too slow to run on every PR. It still runs regularly via the new `nightly-integration.yml` workflow instead.

tests/cli_test.py
aaishwarymishra: I dont think this a problem for the CI, as its cached anyways is you add uv cache in the workflow

.github/workflows/cpu-tests.yml
aaishwarymishra: I would prefer to simply have uv run pytest for normal tests without any flags if possible, a flay or identifier is fine for a specific workflow like the integration tets

aaishwarymishra:
Also, I see copilot commited the code, I am fine with it to be honest, but i will recommend if you take sometime understanding how the github actions and ci works :)
And pytest too if you are not familiar with it .

tests/cli_test.py
aaishwarymishra: Also was the test taking too long or the installation, as I dont think even with cuda build the CI will take 60 mins.

saraelsayed13: It was the installation. `uv sync` was taking over 600 seconds to install all the dependencies, including CUDA, and then timed out. With the CPU-only version, it took less than 200 seconds.

.github/workflows/cpu-tests.yml
saraelsayed13: I just want to clarify your preference: should the integration test be skipped from the normal uv run pytest command and run separately in the nightly workflow using a marker/flag

aaishwarymishra: yes

.github/workflows/nightly-integration.yml
```diff
@@ -0,0 +1,24 @@
+name: nightly integration tests
+
+on:
+  schedule:
+    - cron: "0 3 * * *"
```
aaishwarymishra, line 5: why there is a cron here?

tests/cli_test.py
aaishwarymishra: how did you test it actually? did you ran the actions? If yes can you maybe link it here for me to inspect?

saraelsayed13:
I didn't use Copilot to write the code. I only downloaded it after I made the PR, when I saw the “Download Copilot” button in the repo, just to check it out.

aaishwarymishra:
Yeah , it's ok :)
Don't worry about it

tests/cli_test.py
saraelsayed13: I tested it locally by running the integration test with `pytest -m integration`. The `uv sync` step timed out after 600 seconds while downloading the dependencies. I haven't run the integration workflow on GitHub yet.
--uv run pytest tests/cli_test.py -m integration -v

aaishwarymishra: I think integration test should be CI specific, and CI has higher bandwidth, so I dont think you will hit time limit. You can enable actions on your repo and test your CI there, if you face problem just let me know we can figure something out :)

.github/workflows/nightly-integration.yml
saraelsayed13: I added the nightly workflow based on the discussion in issue #128, where it was mentioned that the slow integration tests should not run on every PR and could instead run nightly. That's why I added the cron here.

aaishwarymishra: I am not actually sure, should the CI be periodic.

saraelsayed13:
Just to confirm, I’ll:

* Remove the CPU-only override.
* Keep the integration test isolated and marked with `integration`.
* Keep it out of the normal `uv run pytest` and run it separately in CI.
* Test the integration workflow on GitHub Actions.
* For the nightly workflow, should I remove it entirely, or keep it with manual triggering only?
* Should I make these changes in this PR or open a new one?

Also, sorry for all the back-and-forth, and thank you for your patience.

aaishwarymishra:
it's fine we can go one step at a time, instead of doing all the changes once do few let  me know I can review , move to the next one.

I am fine with this pr, but if you prefer to break it into separate smaller pr I am fine with it too :)

saraelsayed13:
Removed the CPU-only override

saraelsayed13:
I’ve completed the integration test isolation part. The integration test is now skipped by default with `uv run pytest` and can be run separately with `uv run pytest -m integration`. I’ve pushed the changes for review.

tests/conftest.py
```diff
@@ -16,3 +17,20 @@ def force_dataloader_num_workers_zero():
...
+    regular_tests = []
+
+    for item in items:
+        if "integration" in item.keywords:
+            integration_tests.append(item)
+        else:
+            regular_tests.append(item)
+
+    items[:] = regular_tests
+    config.hook.pytest_deselected(items=integration_tests)
```
aaishwarymishra: ok for what I have read we can define flags for the pytest in the pyproject,toml itself,

we can use the adopts key

aaishwarymishra:
also its good to resolve the comments which you think are fixed :)

tests/conftest.py
saraelsayed13: I had initially implemented it using `addopts`, but I changed it to the `conftest.py` approach because I was concerned about how the default marker would interact with explicitly running `-m integration. After testing it, I confirmed that `addopts` works as expected, so I switched back to it.

saraelsayed13:
I’m still waiting for your thoughts on the nightly workflow. Should I remove it entirely or keep it with manual triggering only?

pyproject.toml
aaishwarymishra, line 1: why this diff is so big even when nothing has changed in many lines?

aaishwarymishra:
> I’m still waiting for your thoughts on the nightly workflow. Should I remove it entirely or keep it with manual triggering only?

I think we can rename it to integration test and it could be same as other tests we have now, I think I noticed in the workflow was you were syncing uv with dev group, I think no dependency should be installed in the root level and only generated project should install dependency it needs to test.

pyproject.toml
saraelsayed13: I think I accidentally deleted the file and then added it back, which caused the large diff. I’ll make sure everything is correct and clean, and if needed, I can open a new clean PR with only the intended changes.
I think I accidentally removed and re-added `pyproject.toml` while editing, which caused the large diff. The only intended changes were in the last few lines of the file.

To avoid any unintended changes, I’ll leave this PR as it is and open a clean follow-up PR with only those intended changes after I make sure everything else is correct.

saraelsayed13:
If the nightly workflow is not essential at this stage, would it be okay if I remove it for now and focus on completing the main requested changes first?

Issue #125, opened 2026-08-04
Add more datasets, support for huggingface datasets, etc


vfdev-5:
Let's add more datasets that user can select from the CLI:
-  [x] generic HF datasets option
-  [ ] built-in HF dataset wrapper for python programming, python-edu, https://huggingface.co/datasets/HuggingFaceTB/smollm-corpus
- built-in HF dataset wrappers for:
  -  [x] https://huggingface.co/datasets/Salesforce/wikitext
  - [ ] https://huggingface.co/datasets/HuggingFaceH4/ultrachat_200k

aaishwarymishra:
ok what should be the scope of these? I get the built in wrappers, but what about the generic datasets how should we handle it?

As we have a `TransformDataset` class which basically takes in datasets and applies the transform if provided. When user is using the hugging face dataset should he provide the download hugging face url? or the local dataset path and it's, her own transform?

atharva7-g:
I'd be interested in contributing to this. 

> ok what should be the scope of these? I get the built in wrappers, but what about the generic datasets how should we handle it?

Usually when pulling datasets from HF, datasets are id'd as: owner/name. So URLs are probably not ideal. And I think we can support local files as well. Parsing should be durable for inferring HF or local. It would also be better if we can add HF as an optional dep here, then we can use the HF API directly for loading datasets.

aaishwarymishra:
Hi @atharva7-g  contributions are welcome, 

whats in my mind is this, 
- For generic wrapper we can take an `__target__` in the the dataset config and have an generic transform which user has to play with a little with inputs to get it working.
- For the case of the specific datasets as listed in the original issue, we will have pre-built transforms for them.

Well I am not sure how code needs to be changed as it's first time we are integrating the external dataset support.

Regarding "HF API directly for loading datasets" I am not sure how to handle it exactly there are multiple ways to load the HF dataset, the easiest way I know of is using the HF datasets package but it loads dataset as there dataset object that then needs to be converted to pytorch dataset which feels a little redundant in my opinion. Also for private datasets requires authentication which requires some HF package I don't remember exactly. Also I am not sure if the HF package should be required dependency or we can keep it optional for projects that require it @vfdev-5  what do you think? I think we can keep it required if its small enough for all the projects.

well I am open for ideas if you want to work on this :)

TahaZahid05:
This sounds good to me. For specific HF datasets, it is doable since we only need to write transforms for them to make them work with Trainite.

[...]

PR #167, opened 2026-09-03
Add WikiText dataset support

esthermichelle:
## Summary

Added built-in support for the `Salesforce/wikitext` dataset.

This includes a dedicated WikiText dataset configuration, transform, registry entry, documentation template, and tests following the existing Hugging Face dataset support pattern.

This PR addresses the WikiText dataset part of #125.

## AI usage

**Tool(s) and usage:**

Claude was used to help me understand the existing dataset architecture and How to implement pattern, and to clarify how to add the WikiText dataset support.

## Testing

- [x] Documentation updated if needed


aaishwarymishra:
btw @esthermichelle do you have worked with hugging face datasets ? or have interest in machine learning?

esthermichelle:
Yes, I have worked with hugging face datasets and I am very interested in machine learning. I've been learning and working on ML-related projects this summer

aaishwarymishra:
Ok, I am not against the AI use, but the issue was intermediate level as it spanned over multiple files, I would suggest you try to read and understand the complete generated code yourself first and then we can continue with the PR :)

trainite/datasets/wikitext.py
```diff
@@ -0,0 +1,78 @@
...
+
+        token_ids = self.tokenizer(
+            text,
+            add_special_tokens=False,
+            truncation=True,
+            max_length=self.max_length - 1,
+        )["input_ids"]
+
+        bos = self.tokenizer.bos_token_id
+        eos = self.tokenizer.eos_token_id
```
aaishwarymishra: Transform for hugging face datasets should not be responsible for the special token addition, tokenizer should do it.

trainite/config/datasets.py
```diff
@@ -142,3 +142,37 @@ class HuggingFaceDataConfig(DataWithAutoSplit):
...
+
+
+class WikiTextDatasetConfig(HuggingFaceDatasetConfig):
+    path: Literal["Salesforce/wikitext"] = "Salesforce/wikitext"
+    name: Literal[
+        "wikitext-103-raw-v1",
+        "wikitext-103-v1",
+        "wikitext-2-raw-v1",
+        "wikitext-2-v1",
+    ] = "wikitext-2-raw-v1"
```
aaishwarymishra: Does this dataset comes with the train, test or validation split on hugging face?

trainite/datasets/wikitext.py
esthermichelle: You're right, I'll update the transform to let the tokenizer handle the special tokens instead of adding BOS/EOS manually

trainite/config/datasets.py
esthermichelle: Yes, the wikitext configurations provide train, validation, and test splits on Hugging Face, and the current configuration uses these splits through the dataset loader.

esthermichelle:
I actually wrote the initial implementation myself, and then used AI to help me review and improve parts of it based on the existing architecture and patterns. Thank you for the feedback.

trainite/config/datasets.py
aaishwarymishra: okay, not really, we have 2 configs for the data config, one in which we split from a dataset like used here, the other where we define separate dataset config for train val etc

esthermichelle: Thanks for the clarification, I understand the distinction now. I updated WikiTextDataConfig to use DataConfigBase with separate train, validation, and test SplitConfig entries, since WikiText provides these splits directly

tests/datasets/wikitext_test.py
```diff
@@ -0,0 +1,87 @@
...
+class FakeTokenizer:
+    def __call__(
+        self,
+        text,
+        add_special_tokens=True,
+        truncation=True,
+        max_length=128,
+    ):
+        assert add_special_tokens is True
+        assert truncation is True
```
aaishwarymishra: why we asserting here?

tests/datasets/wikitext_test.py
aaishwarymishra: why are the tests scope limited to transform? we should also test the dataset itself

trainite/config/datasets.py
```diff
@@ -142,3 +149,65 @@ class HuggingFaceDataConfig(DataWithAutoSplit):
...
```
aaishwarymishra: I dont like this using literals as types, I think many configs use them @TahaZahid05 should we maybe remove them to make code cleaner?

trainite/config/registry.py
```diff
@@ -88,6 +88,17 @@ def dataset_config_cls(self) -> type[BaseModel]:
         preprocessor_spec_name="gpt2",
         dependencies=["datasets"],
     ),
+    "wikitext": DatasetSpec(
+        name="wikitext",
+        implementation_path=Path("trainite/datasets/wikitext.py"),
+        config_cls_path="trainite.config.datasets.WikiTextDataConfig",
+        dataset_config_cls_path="trainite.config.datasets.WikiTextDatasetConfig",
```
aaishwarymishra, line 95: @TahaZahid05 we use path for few paths and str for some, I think it could be a separate issue to make them consistent.

trainite/datasets/wikitext.py
```diff
@@ -0,0 +1,76 @@
...
+        token_ids = tokenized["input_ids"]
+        attention_mask = tokenized["attention_mask"]
+
+        input_ids = torch.tensor(
+            token_ids[:-1],
+            dtype=torch.long,
+        )
+        labels = torch.tensor(
+            token_ids[1:],
+            dtype=torch.long,
```
aaishwarymishra, line 59: why are we shifting the labels shouldn't this be done by collate_fn, I don't exactly remember how other transform of other dataset behave so i might be wrong :)

tests/datasets/wikitext_test.py
esthermichelle: These assertions verify that WikiTextTransform passes the expected arguments to the tokenizer, I can switch this to mock.assert_called_with for consistency if preferred

trainite/datasets/wikitext.py
esthermichelle: The shift is done in the transform following the existing PromptCompletionTransform pattern, while CausalLMCollateFn handles batching and padding, I checked the existing transforms and kept WikiText consistent with that design

tests/datasets/wikitext_test.py
aaishwarymishra: Do we even need them?

trainite/datasets/wikitext.py
aaishwarymishra: @vfdev-5 should the dataset do the shifting ? Now I think about it it should be done by collate_fn tied to model, in that case we would need to have multiple collate_fn for transformer itself, like causal collate_fn

tests/datasets/wikitext_test.py
esthermichelle: Yeah, I'd keep them , I thought they were useful for catching regressions in add_special_tokens and truncation, but I don't mind removing them if you think they're not needed

trainite/datasets/wikitext.py
vfdev-5: This is a datapoint transform, so it is fine like that.

trainite/config/datasets.py
vfdev-5: yes, put str here and in the comment a link to the HF dataset, URL, to know the possible options

vfdev-5:
Checking the PR, I do not understand how to create this wikitext dataset programmatically without the config.

aaishwarymishra:
> Checking the PR, I do not understand how to create this wikitext dataset programmatically without the config.

I am afraid, I did not get what you meant. For the dataset, the approach is to simply use the `dataset.load_dataset` as the `__target__` and pass the parameter like path are then passed to the load_dataset. For the pr in question they simply used the generic  HuggingFaceDatsetConfig  as the parent.

trainite/config/datasets.py
esthermichelle: I'll update it to str and add the Hugging Face dataset URL in the comment, Thank you for the feedback.

trainite/config/registry.py
```diff
@@ -88,6 +88,17 @@ def dataset_config_cls(self) -> type[BaseModel]:
...
+        implementation_symbol="WikiTextTransform",
+        builder_symbol="datasets.load_dataset",
+        readme_template_path=Path("trainite/templates/components/datasets/wikitext.md"),
+        preprocessor_spec_name="gpt2",
+        dependencies=["datasets"],
+    ),
```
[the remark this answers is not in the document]
aaishwarymishra: I think its outdated now,
Edit:that pr to remove the hardcoded parameters on cli generation test is not merged yet, for the formalityI think we can add it for now.

tests/datasets/wikitext_test.py
aaishwarymishra: we dont need it if you pass correct parameters in the test file

trainite/config/registry.py
TahaZahid05: sure i'll open it.

esthermichelle:
Got it, I'll remove the unnecessary assertions and add the wikitext case to the CLI generation test for now, Thanks for the clarification!

trainite/config/registry.py
TahaZahid05: @aaishwarymishra Actually, we have to use Path and str depending on whether we want the template file or the import target. So the use of Path and str is correct. We have to standardize the naming. For instance, when using strs we can have `config_cls_target` so we can differentiate between target and paths. Does that make more sense?

esthermichelle:
@aaishwarymishra could you please review the latest changes, I’ve addressed the review feedback, Thanks for your patience!

trainite/config/registry.py
aaishwarymishra: yeah, it does

esthermichelle: Just to clarify, should I update the naming in this PR? @aaishwarymishra

aaishwarymishra: Not really if should be a separate pr, I will look after this pr Tommorow I hope its fine:)

tests/cli_test.py
```diff
@@ -91,15 +92,23 @@ def test_init_generates_valid_project(models: list[str], dataset: str, trainer:
...
             assert generated_config["data"]["dataset"]["_target_"] == "datasets.load_dataset"
-            assert generated_config["preprocessor"]["_target_"] == ("preprocessors.gpt2_tokenizer.load_gpt2_tokenizer")
+            assert generated_config["preprocessor"]["_target_"] == "preprocessors.gpt2_tokenizer.load_gpt2_tokenizer"
             generated_pyproject = (project_dir / "pyproject.toml").read_text()
             assert "datasets" in generated_pyproject
             assert "transformers" in generated_pyproject
+
         else:
             assert generated_config["data"]["dataset"]["_target_"].startswith("dataset_impl.")
-        assert generated_config["data"]["transform"]["_target_"].startswith("dataset_impl.")
```
aaishwarymishra, line 109: why they are repeated?

tests/cli_test.py
```diff
@@ -91,15 +92,23 @@ def test_init_generates_valid_project(models: list[str], dataset: str, trainer:
             if filename is not None:
                 py_compile.compile(str(project_dir / filename), doraise=True)
```
aaishwarymishra, line 110: this code is almost similar to the hugging face if statement, I think the hugging face datasets should have a common if block, I thing is how to identify them without having to write big if condition to check every hugging face sataset

esthermichelle:
Thanks for pointing this out, I’ll refactor the test to use a common condition for hugging face-backed datasets instead of checking each dataset by name, I’ll also handle the different data split structures and keep the dataset\transform assertions together to avoid the repeated-looking logic, I’ll update the test accordingly

tests/cli_test.py
aaishwarymishra: looks good, but why there are some unnecessary diffs

aaishwarymishra:
Hi, other code looks fine to me, these are some final changes, can you replace the literals with str in pydantic classes, a recent pr got merged that makes all the  literals to str.

And pls take a fresh pull of this branch , I resolved a merge conflict.

And address one more comment I left, and I think this PR can be merged.

Thank you for your help :)

esthermichelle:
Done @aaishwarymishra

aaishwarymishra (approved):

aaishwarymishra:
Thanks

Issue #31, opened 2026-06-01
Add test coverage


aaishwarymishra:
## Description

Currently, our test coverage for the configuration modules (`config`) and the base `registry` needs improvement to ensure long-term stability and prevent regressions.

Additionally, our integration tests currently run directly within the local development Python environment, utilizing already-installed dependencies. This doesn't accurately simulate a clean, end-user installation. When a TOML file and project dependencies are generated dynamically, we should programmatically isolate the test execution environment. Specifically, we should use `uv` to install the newly generated dependencies and run the tests against that isolated environment.

### 1. Unit Testing & Coverage

* [x] **Expand `config` tests:** Add edge-case testing for configuration loading, missing fields, invalid TOML/YAML formats, and environment variable overrides.
* [x] **Expand `registry` tests:** Add unit tests for base registry registration, lookup failures, duplicate keys, and dynamic loading mechanics.

### 2. Test Environment Isolation (Integration Tests)

* [ ] **Introduce `uv` environment creation:** Modify the integration test setup (or specific fixtures) to spin up a temporary, clean virtual environment using `uv venv`.
* [ ] **Dynamic Dependency Installation:** For tests involving generated TOML files, use `uv pip install` to explicitly install those generated dependencies into the temporary environment.
* [ ] **Test Execution Shift:** Ensure the integration test suite runs commands/subprocesses targeting the isolated environment's Python binary rather than the active development environment.

webdevsamran:
Hi team, I would like to work on this issue if it's available for assignment.

I can tackle the testing roadmap outlined here:
1. Adding unit test coverage for `config` (schema edge cases, missing fields, environment variable overrides) and base `registry` mechanics (duplicate registration handling, lookup failures, dynamic loading).
2. Updating the integration test fixtures to programmatically spin up isolated test environments via `uv venv` and install dynamically generated dependencies with `uv pip install`, ensuring integration tests run against clean, isolated virtual environments rather than the host dev Python environment.

Could you please assign this to me? Thanks!

aaishwarymishra:
I don't think this issue is open right now as its old and some of the item s listed there are outdated.

aaishwarymishra:
@webdevsamran well there is one, there are multiple pydantic classes in trainite/config which uses literal, I think they can be turned into simple str. You can decide whether to use literal or str, but most of the time literal makes it unnecessarly verbose, PR is welcomed :)

PR #173, opened 2026-09-07
refactor(config): simplify model and dataset target types to str

webdevsamran:
Follow up on discussion in #31.

### Description

Currently, several Pydantic configuration classes in `trainite/config/models.py` and `trainite/config/datasets.py` type their `target` (`_target_`) fields as multi-line `Literal[...]` unions duplicating full import paths (`"trainite.models..."` vs `"models..."`). This makes the config definitions unnecessarily verbose and rejects custom user-provided implementations with strict validation errors.

Following maintainer guidance (@aaishwarymishra in #31), this PR:
1. Replaces `Literal[...]` on `target` fields across `trainite/config/models.py` and `trainite/config/datasets.py` with `str`.
2. Preserves default values pointing to canonical module paths.
3. Adds unit test coverage in `tests/config/base_test.py` ensuring all model and dataset configs instantiate cleanly with string targets.

### DIFF
--- a/trainite/config/datasets.py
+++ b/trainite/config/datasets.py
@@ -1,13 +1,10 @@
-from typing import Literal, Self
+from typing import Self
 from pydantic import Field, model_validator
 from trainite.config.base import DatasetConfig, TransformConfig, DataWithAutoSplit, DataLoaderConfig
 
 
 class PromptCompletionTransformConfig(TransformConfig):
-    target: Literal[
-        "trainite.datasets.string_reverse.PromptCompletionTransform",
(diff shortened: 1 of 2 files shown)


tests/config/base_test.py
aaishwarymishra, line 1: I dont think we need test for this

webdevsamran:
Thanks for the feedback @aaishwarymishra! Removed the test additions as requested - the PR now only touches `trainite/config/datasets.py` and `trainite/config/models.py` (2 files, 9 insertions / 32 deletions). Force-pushed to the same branch.

aaishwarymishra:
Thanks :)

PR #174, opened 2026-09-07
Add integration test workflow

saraelsayed13:
## Summary

Add a separate GitHub Actions workflow for integration tests. The workflow runs automatically on pushes and pull requests to `main` and `develop`, while keeping integration tests separate from the regular CPU test workflow.

The integration test was also adjusted to use a smaller model configuration to reduce the test workload and CI runtime.

Relates to #128
## AI usage

* [x] No AI tools were used for this contribution
* [ ] AI tools were used to assist this contribution

## Testing

* [x] Tests added or updated where needed
* [x] Existing tests and checks pass
* [ ] Documentation updated if needed


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: actions/checkout@v7
+
+      - name: Install uv and Python
+        uses: astral-sh/setup-uv@v9.0.0
+        with:
+          enable-cache: true
+          python-version: "3.11"
+
+      - name: Sync dependencies
+        run: uv sync --locked --dev
```
aaishwarymishra: I think we should only install deps required to generate and run tests, as deps should be installed by test itself

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+      - "tests/**"
+      - "pyproject.toml"
+      - "uv.lock"
+      - ".github/workflows/**"
+  workflow_dispatch:
+
+jobs:
+  integration-tests:
+    name: Integration tests
+    runs-on: ubuntu-latest
```
aaishwarymishra: It should run on linux, mac and windows like non integration tests run

.github/workflows/integration-tests.yml
saraelsayed13: Like this:

`uv run --with pytest --no-group dev pytest -m integration`

aaishwarymishra: yeah

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: actions/checkout@v7
+
+      - name: Install uv and Python
+        uses: astral-sh/setup-uv@v9.0.0
+        with:
+          enable-cache: true
+          python-version: ${{ matrix.python-version }}
+
+      - name: Run integration tests
+        run:  uv run --with pytest --with torch --no-group dev pytest -m integration
```
aaishwarymishra:
```suggestion
        run:  uv run --with pytest --with torch pytest -o addopts="" -m integration
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: actions/checkout@v7
+
+      - name: Install uv and Python
+        uses: astral-sh/setup-uv@v9.0.0
+        with:
+          enable-cache: true
+          python-version: ${{ matrix.python-version }}
+
+      - name: Run integration tests
+        run: uv run --with pytest --with torch --with transformers --no-group dev pytest -m integration
```
aaishwarymishra:
```suggestion
   uv run --with pytest --with torch pytest -o addopts="" -m integration
```

aaishwarymishra:
Thanks :)

PR #175, opened 2026-09-10
test: remove dependency on test init files

### LINKED ISSUE #158: Remove unnecessary `__init__.py` files from `tests/`

esthermichelle:
## Summary

This PR builds on top of #159 to resolve #158.

It removes the dependency on `__init__.py` files in the tests by passing test classes and functions directly instead of using hardcoded module paths.

A helper was added to resolve the import path from `__module__` and `__qualname__`, so the tests can still dynamically locate the required targets without relying on the test directories being Python packages.

This keeps the test code readable while allowing the `__init__.py` files to be removed from the test directories.

## AI usage

- [x] No AI tools were used for this contribution
- [ ] AI tools were used to assist this contribution

**Tool(s) and usage:**

## Testing


esthermichelle:
Hi @TahaZahid05, the PR is ready for review

TahaZahid05:
@Mohammed-Adel1 would you like to review the PR? It can help you in understanding how the problem was solved and potentially point out any issues while reviewing. You can let me know once you have reviewed and I can then take a look.

Mohammed-Adel1 (approved):
That's a great fix and sounds good to me .
For context on why this matters: without __init__.py, pytest's default import mode (prepend) collects this test file as a top level module (decoder_trainer_test), while the old hardcoded strings ("tests.trainers.decoder_trainer_test.SimpleModel") assumed the fully qualified package path. That mismatch caused the same file to be imported twice under two different module names, and each import created a distinct copy of classes like SimpleModel which is why isinstance() checks and is comparisons on collate_fn were failing in CI .
Resolving the target dynamically via f"{obj.__module__}.{obj.__qualname__}" sidesteps this nicely since it always reflects whatever module name Python actually used at runtime regardless of __init__.py presence or pytest's import mode.

TahaZahid05 (approved):

PR #177, opened 2026-09-12
feat: add Python-Edu dataset

saraelsayed13:
## Summary

Adds a built-in `python-edu` dataset wrapper for `HuggingFaceTB/smollm-corpus` .
(part of #125).

`python-edu` provides file metadata only (`blob_id`, `score`, `int_score`, etc.), without the actual source code. File contents are fetched separately from Software Heritage's public S3 bucket and gzip-decoded, so a dedicated `PythonEduDataset` is used to download the files before training.

* `min_int_score` filters samples by the HuggingFaceTB classifier score (default: `4`).
* `max_samples` enables lazy streaming for small local tests, avoiding the need to load the full metadata split.
* Failed downloads are automatically filtered out.

## AI usage

* [x] AI tools were used to assist this contribution

## Testing

* [x] Tests added or updated where needed
* [x] Existing tests and checks pass
* [x] Documentation updated if needed

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -41,4 +41,4 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Run integration tests
-        run: uv run --with pytest --with datasets pytest -o addopts="" -m integration tests/cli_test.py
+        run: uv run --with pytest  pytest -o addopts="" -m integration tests/cli_test.py
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -41,4 +41,4 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Run integration tests
-        run: uv run --with pytest pytest -o addopts="" -m integration tests/cli_test.py
+        run: uv run --with pytest --with datasets pytest -o addopts="" -m integration tests/cli_test.py
--- a/trainite/config/datasets.py
+++ b/trainite/config/datasets.py
@@ -189,3 +189,29 @@ class WikiTextDataConfig(DataConfigBase):
             ),
         )
     )
+
+
+class PythonEduTransformConfig(TransformConfig):
+    target: str = Field(
+        default="trainite.datasets.python_edu.PythonEduTransform",
+        alias="_target_",
+    )
+    max_length: int = Field(default=128, gt=1)
+
+
+class PythonEduDatasetConfig(DatasetConfig):
+    target: str = Field(
+        default="trainite.datasets.python_edu.PythonEduDataset",
+        alias="_target_",
+    )
+    split: str = Field(default="train", min_length=1)
+    min_int_score: int = Field(default=4, ge=0, le=5)
+    max_samples: int | None = None
+
+
+class PythonEduDataConfig(DataWithAutoSplit):
+    dataset: PythonEduDatasetConfig = Field(default_factory=PythonEduDatasetConfig)
+    transform: PythonEduTransformConfig = Field(default_factory=PythonEduTransformConfig)
+    test_ratio: float = 0.1
+    val_ratio: float = 0.1
+    dataloader: DataLoaderConfig = Field(default_factory=lambda: DataLoaderConfig(batch_size=32, shuffle=True))
--- a/trainite/datasets/__init__.py
+++ b/trainite/datasets/__init__.py
@@ -6,12 +6,10 @@ from trainite.datasets.counting import (
 )
 from trainite.datasets.hugging_face import HuggingFaceTransform
 from trainite.datasets.wikitext import WikiTextTransform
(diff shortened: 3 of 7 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -41,4 +41,4 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Run integration tests
-        run: uv run --with pytest pytest -o addopts="" -m integration tests/cli_test.py
+        run: uv run --with pytest --with datasets pytest -o addopts="" -m integration tests/cli_test.py
```
aaishwarymishra: why we need this ?

trainite/datasets/python_edu.py
aaishwarymishra: I dont think we need to do all this, you can use wikitext dataset as reference how its implemented

trainite/datasets/__init__.py
aaishwarymishra, line 1: @TahaZahid05 do we really need to add the modules to __init__ files, as they cause imports error consistently as they import everything even when we dont need them, we can simply use complete path like `trainite.dataset.module` I dont think its a problem

TahaZahid05: That is true.. The end-user never has a use case for the init files, and neither does it help in the development. we can remove it.

trainite/datasets/python_edu.py
aaishwarymishra, line 1: @vfdev-5 Do you think we should use this dataset, as we need to download the individual blob ourselves?

Also downloading individual blob be slow might have to add parallel download but we might get rate limited.

Also should we download the whole dataset at once and save it in a folder to avoid re-downloading? but it will be slow as dataset is big.

Another option is to download only the number of rows and criteria we need, but we have to re-download the dataset on every run as there wont be a way to very new dataset requested was same as before.

vfdev-5: This url explains how to fetch the data: https://huggingface.co/datasets/HuggingFaceTB/smollm-corpus#python-edu
However, we need to have some creds to be setup in order to be able to get access to the S3...
If we can find any other easily accessible python code datasets of reasonable size ~1GB we can use it instead of this dataset.

aaishwarymishra: Well, I ran this branch on my machine but I added code to stop when required dataset size reached, for dataset len 10, it worked but I am not sure if it will work with big dataset as they themselves recommend using boto, so yeah we can look for alternatives I suppose.

trainite/config/datasets.py
```diff
@@ -189,3 +189,39 @@ class WikiTextDataConfig(DataConfigBase):
...
+    train: SplitConfig = Field(
+        default_factory=lambda: SplitConfig(
+            dataset=PythonEduDatasetConfig(split="train[:90%]"),
+            transform=PythonEduTransformConfig(),
+            dataloader=DataLoaderConfig(batch_size=32, shuffle=True),
+        )
+    )
+    val: SplitConfig = Field(
+        default_factory=lambda: SplitConfig(
+            dataset=PythonEduDatasetConfig(split="train[90%:]"),
```
aaishwarymishra: is there no validation split?

trainite/templates/components/datasets/python_edu.md
aaishwarymishra, line 1: I dont like the readme, we dont need to list the differences, explain about dataset and how to use it, and reference which dataset it is subset of

.github/workflows/integration-tests.yml
```diff
@@ -41,4 +41,4 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Run integration tests
-        run: uv run --with pytest pytest -o addopts="" -m integration tests/cli_test.py
+        run: uv run --with pytest  pytest -o addopts="" -m integration tests/cli_test.py
```
aaishwarymishra: fix this unnecessary diff

trainite/config/datasets.py
saraelsayed13: yes, only train and test exist

trainite/config/datasets.py
```diff
@@ -189,3 +189,39 @@ class WikiTextDataConfig(DataConfigBase):
...
+class PythonEduTransformConfig(TransformConfig):
+    target: str = Field(
+        default="trainite.datasets.python_edu.PythonEduTransform",
+        alias="_target_",
+    )
+    max_length: int = Field(default=128, gt=1)
+
+
+class PythonEduDatasetConfig(HuggingFaceDatasetConfig):
+    path: str = "tyoc213/split-avelina-python-edu-distilled"
```
aaishwarymishra: Hi, can you maybe use https://huggingface.co/datasets/Avelina/python-edu-cleaned the 66k dataset we checked is kind of too low maybe.

aaishwarymishra (approved):

aaishwarymishra:
LGTM thanks !

PR #183, opened 2026-09-15
Gemma moe

aaishwarymishra:
## Summary

Adds support for the Gemma 4 moe

## AI usage

**Tool(s) and usage:**

Codex for helping with writing code and debugging

## Testing

- [ ] Tests added or updated where needed
- [ ] Existing tests and checks pass


trainite/models/gemma4_moe.py
```diff
@@ -0,0 +1,565 @@
...
+            if attention_mask.ndim == 2:
+                allowed = allowed & attention_mask[:, None, :].bool()
+            elif attention_mask.ndim == 3:
+                allowed = allowed & attention_mask.bool()
+            else:
+                raise ValueError("attention_mask must have 2 or 3 dimensions")
+        if self.sliding_window is not None:
+            allowed = allowed & create_sliding_mask(position_ids, self.sliding_window)
+
+        attended = F.scaled_dot_product_attention(
```
TahaZahid05, line 293: shouldn't we keep `dropout` as configurable as well in the class?

trainite/models/gemma4_moe.py
```diff
@@ -0,0 +1,565 @@
...
+            num_attention_heads,
+            num_key_value_heads,
+            head_dim=head_dim,
+            rope_theta=rope_theta,
+            rope_scaling_factor=rope_scaling_factor,
+            rotary_fraction=rotary_fraction,
+            sliding_window=sliding_window,
+            key_equals_value=key_equals_value,
+        )
+        self.pre_attention_norm = nn.RMSNorm(hidden_size, eps=1e-6)
```
TahaZahid05, line 338: same for `eps`? should we have configurable like `rms_norm_eps`?

trainite/config/models.py
```diff
@@ -25,3 +25,23 @@ class RoPETransformerModelConfig(ModelConfig):
...
+    hidden_size: int = Field(default=64, gt=0)
+    num_layers: int = Field(default=2, gt=0)
+    num_attention_heads: int = Field(default=4, gt=0)
+    num_key_value_heads: int = Field(default=2, gt=0)
+    dense_intermediate_size: int = Field(default=128, gt=0)
+    expert_dim: int = Field(default=64, gt=0)
+    num_experts: int = Field(default=4, gt=0)
+    top_k: int = Field(default=2, gt=0)
+    head_dim: int = Field(default=16, gt=0)
+    layer_types: tuple[str, ...] = ("sliding", "global")
```
TahaZahid05: This seems painful from the user's perspective. For instance, if user has a model with more layers, they have to define a list in the config where they tell for each layer, whether it is sliding and global. This can be too long, and also error-prone.

In my implementation, I used `sliding_ratio: int = 5` (5 sliding layers followed by 1 global layer, with the last layer always being global). This way, they only need to define the pattern and we can handle assigning what type of attention to use in each layer.

I think we can have both where `layer_types` can be used by user's for more custom cases while in general they can use `sliding_ratio`.

trainite/models/gemma4_moe.py
aaishwarymishra: Hmm, in the jax implementation they didnt use dropout anywhere and only place where dropout was used in Hugging face implementation was in attention block.

trainite/models/gemma4_moe.py
aaishwarymishra: Not sure actually, I think keeping configuration simple is better than giving user every knob to tweak.

trainite/config/models.py
aaishwarymishra: well true but the user may want some other pattern?

TahaZahid05: That is my point. If user want to use a custom pattern, they can use `layer_types` if provided by user else we will use `sliding_ratio` value. Point is that for majority cases, `sliding_ratio` works. `layer_types` can be used incase of a custom pattern.

trainite/models/gemma4_moe.py
TahaZahid05: I think let's add dropout since it is quite common to add it?

trainite/config/models.py
aaishwarymishra: Not sure lets discuss in meeting

aaishwarymishra:
I think we could rename the model file, and have a common test for both moe and dense as both essentially have same components, and have a common readme as they are in same file.

I would also recommend adding an weight loading test similar to tinyrandom/moe to check if weights can be loaded correctly

PR #188, opened 2026-09-17
fix(init): add explicit build-system to generated project pyproject.toml

### LINKED ISSUE #182: Generated experiment fails on `pip install -e .`
When following the Quickstart instructions in the README to initialize and set up a generated experiment project, running `pip install -e .` fails with a Setuptools error regarding multiple top-level packages.

This blocks new users and contributors setting up a generated project using standard `pip` and Python virtual environments.

## Steps to Reproduce

1. Generate a starter experiment as instructed in the README Quickstart:
   ```bash
   trainite init my-experiment --model rope-transformer --dataset string-reverse
   ```
2. Navigate into the generated experiment:
   ```bash
   cd my-experiment
   ```
3. Set up a virtual environment using Python's standard `venv`:
   ```bash
   python3 -m venv .venv
   source .venv/bin/activate
   ```
4. Run the dependency installation step from the README:
   ```bash
   pip install -e .
   ```

## Error Output

```bash
pip install -e .
Looking in indexes: https://pypi.org/simple, https://packagecloud.io/github/git-lfs/pypi/simple
Obtaining file:///home/taha/Desktop/trainite/trainite/my-cool-experiment
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... error
  error: subprocess-exited-with-error

  × Getting requirements to build editable did not run successfully.
  │ exit code: 1
  ╰─> [14 lines of output]
      error: Multiple top-level packages discovered in a flat-layout: ['models', 'dataset_impl', 'preprocessors'].

      To avoid accidental inclusion of unwanted files or directories,
      setuptools will not proceed with this build.

      If you are trying to create a single distribution with multiple packages
      on purpose, you should not rely on automatic discovery.
      Instead, consider the following options:

      1. set up custom discovery (`find` directive with `include` or `exclude`)
      2. use a `src-layout`
      3. explicitly set `py_modules` or `packages` with a list of names

      To find more information, look for "package discovery" on setuptools docs.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build editable did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

We need to implement a fix so that users following the README setup with `pip` can install dependencies without any errors.

ahmed-226:
## Summary

Closes #182. 
Following the README Quickstart, `pip install -e .` inside a generated experiment fails with:

    error: Multiple top-level packages discovered in a flat-layout: ['models', 'dataset_impl', 'preprocessors']

Root cause: the generated `pyproject.toml` had no `[build-system]`, so pip defaults to setuptools, whose auto-discovery rejects the flat layout with multiple top-level packages.

Fix: `generate_uv_project()` now emits an explicit `[build-system]` (`setuptools.build_meta`) and `[tool.setuptools] packages = []`, which is the remedy the setuptools error itself suggests. `pip install -e .` now succeeds and installs the project's dependencies.

Notes:
- Hatchling was tried first to mirror the parent repo's backend, but its wheel builder requires a single top-level package matching the project name (`default_file_selection_options`), which a flat script layout cannot satisfy.

## AI usage

**Tool(s) and usage:**
opencode — explore the root cause and drafted this PR description.

## Testing

- [x] Tests added: `tests/cli_test.py::test_generated_pyproject_has_build_system`
- [x] `pytest -m "not integration"` → 130 passed
- [x] `pre-commit run --all-files` → all passed (ruff + pyrefly)
- [x] Manual repro: generated a project via `python -m trainite.cli init demo ...`, then `pip install -e .` succeeded end-to-end (metadata, editable wheel, deps, `Successfully installed demo-0.1.0`).


TahaZahid05 (commented):
The fix looks good! However, since through `pip install -e .` we are skipping making an actual package, we need to add a note inside project's generated `README` for `pip install -e .` step that currently the command only installs the project dependencies rather than making a package, and that user has to modify the code if they want to make it a installable package and how they can do it.

ahmed-226:
> The fix looks good! However, since through `pip install -e .` we are skipping making an actual package, we need to add a note inside project's generated `README` for `pip install -e .` step that currently the command only installs the project dependencies rather than making a package, and that user has to modify the code if they want to make it a installable package and how they can do it.

@TahaZahid05 I added a > [!NOTE] under the pip install -e . step in the generated project's README template (trainite/templates/project/README.md)  explains that pip install -e . currently only installs the project's dependencies (the project isn't built as an installable package and runs from source via python main.py config.yaml), and documents how to make it an installable package if desired (src layout + [tool.setuptools.packages.find]).

TahaZahid05 (approved):
Thanks!

PR #189, opened 2026-09-18
fix: make CLI defaults independent of registry ordering

### LINKED ISSUE #185: Make CLI defaults independent of registry ordering
## Problem

Some interactive CLI defaults are selected through the first registry entry, for example:

```python
DATASET_CHOICES[0]
TRAINER_CHOICES[0]
```

This makes behavior depend on dictionary insertion order. Adding or reordering a registry entry can silently change the default shown to users and alter generated projects.

Defaults are product choices, while registry order is an implementation detail. They should not be coupled.

## Expected behavior

Interactive and non-interactive defaults should remain stable when registry entries are reordered.

esthermichelle:
## Summary 
 
 

This PR makes the CLI defaults explicit and independent of the ordering of the registered models, datasets, and trainers to closes #185.

Previously, the interactive CLI selected some defaults using the first item in the corresponding registry/choices list. This made the effective default dependent on registry ordering, so changing the order of registered items could unintentionally change the CLI behavior.

This PR:
- Defines explicit `DEFAULT_MODEL`, `DEFAULT_DATASET`, and `DEFAULT_TRAINER` constants in the registry.
- Updates the interactive CLI to use these explicit constants instead of relying on the first item in the choices lists.
- Updates the non-interactive `Init` configuration defaults to use the same explicit constants.
- Adds regression coverage to verify that interactive defaults remain correct even when the dataset registry ordering is changed.
- Adds coverage for the default values of the `Init` configuration.
 
## AI usage 
 
 
 
**Tool(s) and usage:** 
 
 

Claude was used for code review.
 
## Testing 
 
 
 
### Tests added

The CLI tests were updated to verify that the defaults are explicit rather than derived from registry ordering.

1. **Interactive CLI defaults**
   - The dataset choices are temporarily reversed during the test.
   - This ensures that the first registered dataset is different from `DEFAULT_DATASET`.
   - The test captures the defaults passed to the interactive prompts.
   - It verifies that the model, dataset, and trainer prompts use `DEFAULT_MODEL`, `DEFAULT_DATASET`, and `DEFAULT_TRAINER` respectively.
   - This prevents future changes to registry ordering from silently changing the CLI defaults.

2. **Non-interactive `Init` defaults**
   - A new test creates an `Init` configuration using its default values.
   - It verifies that the model, dataset, and trainer defaults match the explicit registry constants.


tests/cli_test.py
TahaZahid05, line 1: Add a test to ensure the default constant also exist in registry

TahaZahid05:
@esthermichelle Also resolve the conflict please.

TahaZahid05 (approved):
Looks good to me!

PR #191, opened 2026-09-19
feat: add grid search parameter sweeping to project generator

Arun-sunny23:
Description:
This PR introduces native grid search capabilities to the counting example, allowing users to automatically run multiple training iterations across a matrix of hyperparameters defined directly in their config.yaml.
Key Changes:
⚬	Added load_grid_configs in utils.py: Parses the sweep block from the configuration file. It utilizes itertools.product to generate all possible combinatorial variations of the specified parameters. If no sweep block is present, it defaults to returning a single configuration, ensuring full backward compatibility.
⚬	Refactored main.py entry point: Updated the execution flow to accept a list of parsed configurations and sequentially instantiate and run the PyTorch-Ignite Trainer for each configuration in the grid.
Motivation:
Previously, experimenting with multiple parameters required manual updates or external orchestration scripts. This update streamlines hyperparameter tuning by keeping the logic entirely within the standalone example's existing YAML configuration and utility structures.
Testing:
⚬	Verified the script successfully detects multi-variable sweep blocks.
⚬	Confirmed the fallback behavior works smoothly for single-run configurations.
⚬	Tested local execution successfully with Tensorboard logging.


Arun-sunny23:
"Hey @aaishwarymishra! I've put together a lightweight grid-search implementation for the counting example (built with itertools.product to avoid Hydra complexity). The Pull Request is up and all automated checks are passing successfully! Would love your thoughts when you have a chance."

aaishwarymishra:
well the checks are passing because you added it an example so no tests were run, I would prefer if you add a test and the actual implementation so it can be tested on generated code.

Arun-sunny23:
Got it, thanks for the clarification @aaishwarymishra! That makes complete sense regarding the CI coverage.

I'll move the sweep logic into the core codebase so it integrates directly with the generator, and add corresponding unit/integration tests under `tests/` to validate it against generated code. 

Before I push the refactor: do you have a preferred location/module in mind for where you'd like the grid search logic and configuration options to live, or should I follow the convention used by the existing runners?

Arun-sunny23:
Thanks for pointing that out, @aaishwarymishra! That makes complete sense regarding the CI checks not running against the example folder[span_0](start_span)[span_0](end_span).

I'll move the grid search parameter sweeping logic into the actual generator templates and add tests under `tests/` to validate the logic directly against generated code[span_1](start_span)[span_1](end_span). I will push an update to this PR once the tests and template integration are ready[span_2](start_span)[span_2](end_span).

Arun-sunny23:
All requested updates have been implemented and verified:

- **Shared implementation**: Moved `load_grid_configs` into `trainite/shared/utils.py` and updated `trainite/shared/main.py` so projects scaffolded via `trainite init` natively support parameter sweeps.
- **Unit test coverage**: Added unit tests in `tests/utils_test.py` verifying single config passthroughs as well as Cartesian product combinations for multi-parameter grids.
- **Generator test coverage**: Added an integration test in `tests/cli_test.py` ensuring the generated `utils.py` and `main.py` files contain the expected sweep logic.
- **Code style & CI**: Applied `ruff` formatting, rebased against latest `main`, and confirmed all CI checks and test matrices are passing.

Ready for review!

TahaZahid05:
@Arun-sunny23 There is no need to update existing examples with new features, so kindly remove those changes. As for the PR, can you provide some steps on how I can test this by creating an experiment? Currently, it seems unclear how a new experiment provided grid search parameter support.

Arun-sunny23:
Hey @TahaZahid05, good catch on keeping the existing examples untouched — totally agree! I've reverted the changes under `examples/`, so this PR is now strictly focused on the generator templates (`trainite/shared/`) and tests.

Here’s a quick walkthrough on how to test the sweep functionality on a fresh experiment:

### 1. Scaffold a test project
```bash
trainite init test-sweep --model rope-transformer --dataset counting --trainer decoder-trainer
cd test-sweep
```

### 2. Add a `sweep` block to `config.yaml`
Open up `config.yaml` and add the parameters you want to sweep at the bottom:

```yaml
sweep:
  optimizer.lr: [0.001, 0.0005]
  data.train.dataloader.batch_size: [16, 32]
```

### 3. Run it
```bash
python main.py
```

It will detect the `sweep` block, generate the Cartesian product (4 runs in this case), and run through each configuration sequentially:

```text
=== Starting Grid Search Run 1 of 4 ===
...
=== Starting Grid Search Run 2 of 4 ===
...
```

If the `sweep` key is omitted, it behaves normally and runs a single training pass as before.

Let me know if this works or if you'd like anything else adjusted!

Arun-sunny23:
Hi @aaishwarymishra, just following up — the PR should be ready whenever you get a chance to review the final merge. Thanks!

trainite/shared/utils.py
```diff
@@ -134,6 +135,35 @@ def load_config(path: str | Path, config_cls: type[T]) -> T:
...
+
+    # Generate all combinations
+    combinations = list(itertools.product(*values))
+
+    # Create and validate a distinct configuration for each combination
+    configs = []
+    for combo in combinations:
+        run_conf = raw_conf.copy()
+        for key, val in zip(keys, combo):
+            OmegaConf.update(run_conf, key, val)
```
TahaZahid05, line 160: what if user makes a typo in writing keys for the sweep?  We should check if the key even exists before running sweep on it and raising a error to inform user.

trainite/shared/main.py
TahaZahid05, line 18: what about run names? How will user know which run directory corresponds to which run in the sweep? It will be better if the run name describes that.

TahaZahid05, line 18: For the config saved for a particular run, does it's yaml contain the sweep block or no? I think for a particular run's yaml, we can remove the sweep block since it is not relevant for us to see details about sweep for a particular run

trainite/shared/main.py
```diff
@@ -1,17 +1,21 @@
...
     parser.add_argument("config", nargs="?", default="config.yaml")
     args = parser.parse_args()
-    config = load_config(Path(args.config), ProjectConfig)
-    trainer = Trainer(config)
-    trainer.run()
+
+    configs = load_grid_configs(Path(args.config), ProjectConfig)
+    for i, config in enumerate(configs):
+        if len(configs) > 1:
+            print(f"\n=== Starting Grid Search Run {i + 1} of {len(configs)} ===")
```
TahaZahid05, line 16: Printing the active parameters can help user be informed which specific combination is being used for that search.

tests/utils_test.py
```diff
@@ -63,3 +63,44 @@ def test_instantiate_kwargs_override():
...
+# Grid Search Tests
+# ==========================================
+
+from trainite.shared.utils import load_grid_configs
+
+
+class MockSweepConfig(BaseModel):
+    lr: float = 0.001
+    batch_size: int = 32
+    notes: str = "default"
```
TahaZahid05, line 78: Trainite's yaml is nested. So it is:
```yaml
optimizer:
  lr: 0.001
```

sweep should also follow this format rather than current one `optimizer.lr: [0.001, 0.0005]`

TahaZahid05 (commented):
@Arun-sunny23 Thanks. I have left review. Please also add relevant documentation. Also update PR description.

Arun-sunny23:
"@TahaZahid05 Updated! Let me know if it's good to merge."

trainite/shared/utils.py
```diff
@@ -134,6 +135,63 @@ def load_config(path: str | Path, config_cls: type[T]) -> T:
...
+    # Create and validate a distinct configuration for each combination
+    configs = []
+    for combo in combinations:
+        run_conf = raw_conf.copy()
+
+        # TASK 3: Remove the sweep block since it is no longer relevant for the individual run
+        if "sweep" in run_conf:
+            del run_conf["sweep"]
+
+        for key, val in zip(keys, combo):
```
TahaZahid05: why are we checking whether key exist inside the loop where we are running the loop for combinations? This should run once upfront. Also, does `OmegaConf.select` work if the value set for the key is set to null?

trainite/shared/main.py
```diff
@@ -1,17 +1,48 @@
...
+    # Re-read the sweep block to get the keys and combinations for logging and folder naming
+    raw_conf = OmegaConf.load(args.config)
+    sweep_params = raw_conf.get("sweep", None)
+
+    if sweep_params and len(configs) > 1:
+        # Flatten the dictionary to match the itertools output order from utils.py
+        flat_sweep = flatten_sweep_config(OmegaConf.to_container(sweep_params, resolve=True))
+        keys = list(flat_sweep.keys())
+        values = [v if isinstance(v, list) else [v] for v in flat_sweep.values()]
+        combinations = list(itertools.product(*values))
```
TahaZahid05: This seems like a bad way to do this. Why are we repeating logic already written inside `load_grid_configs`?

trainite/shared/utils.py
```diff
@@ -134,6 +135,63 @@ def load_config(path: str | Path, config_cls: type[T]) -> T:
...
+    Flattens a nested dictionary into dot notation.
+    Example: {'optimizer': {'lr': [0.01, 0.001]}} -> {'optimizer.lr': [0.01, 0.001]}
+    """
+    items = []
+    for k, v in d.items():
+        new_key = f"{parent_key}{sep}{k}" if parent_key else k
+        if isinstance(v, dict) or type(v).__name__ == "DictConfig":
+            items.extend(flatten_sweep_config(v, new_key, sep=sep).items())
+        else:
+            items.append((new_key, v))
```
TahaZahid05: what's the need for this else conditional?

trainite/shared/utils.py
```diff
@@ -134,6 +135,63 @@ def load_config(path: str | Path, config_cls: type[T]) -> T:
...
+    sweep_params = raw_conf.get("sweep", None)
+
+    # If there is no sweep block, return the single config inside a list
+    if not sweep_params:
+        return [config_cls.model_validate(raw_conf)]
+
+    # Extract and format the grid parameters
+    sweep_params_dict = OmegaConf.to_container(sweep_params, resolve=True)
+
+    # TASK 1: Flatten the nested dictionary into dot-notation keys
```
TahaZahid05: Please remove unnecessary comments.

Arun-sunny23:
"@TahaZahid05 I've pushed the requested changes based on your feedback and cleaned up the CI formatting issues. Ready for another look when you have time."

